PR #4, opened 2025-11-12
KVStore::remove Implementation Part 2

vidyasilai:
This PR contains the tree algorithm code related to deletions:

- Node and leaf merging
- New `InMemoryNode` update buffer level type `HybridLevel`


src/turtle_kv/core/testing/generate.hpp
```diff
@@ -184,8 +184,11 @@ class RandomResultSetGenerator : public MinMaxSize<usize{1} << 24>
   }
 
   template <bool kDecayToItems, typename Rng>
-  MergeCompactor::ResultSet</*kDecayToItems=*/kDecayToItems>
-  operator()(DecayToItem<kDecayToItems>, Rng& rng, llfs::StableStringStore& store)
+  MergeCompactor::ResultSet</*kDecayToItems=*/kDecayToItems> operator()(
+      DecayToItem<kDecayToItems>,
+      Rng& rng,
+      llfs::StableStringStore& store,
+      Optional<std::vector<KeyView>> to_delete = None)
```
tonyastolfi: nit: this can just be a `std::vector` (no need for `Optional`)

src/turtle_kv/kv_store_scanner.cpp
```diff
@@ -435,15 +435,20 @@ Status KVStoreScanner::set_next_item()
...
-      if (!this->keys_only_ && this->next_item_->needs_combine()) {
+      if (this->next_item_->needs_combine()) {
         this->next_item_->value = combine(this->next_item_->value, scan_level->value());
       }
 
     } else {
-      break;
+      if (this->next_item_->value == ValueView::deleted()) {
+        this->next_item_ = None;
+        continue;
```
tonyastolfi, line 504: Please add comment here about why this is correct.

src/turtle_kv/kv_store_scanner.cpp
```diff
@@ -454,7 +459,8 @@ Status KVStoreScanner::set_next_item()
       LatencyTimer timer{batt::Every2ToTheConst<8>{},
                          KVStoreScanner::metrics().heap_remove_latency};
       this->heap_.remove_first();
-      this->needs_resume_ = true;
+      //this->needs_resume_ = true;
```
tonyastolfi: Remove dead code?

src/turtle_kv/core/testing/generate.hpp
vidyasilai: Done

src/turtle_kv/kv_store_scanner.cpp
vidyasilai: Done

src/turtle_kv/core/merge_compactor.cpp
```diff
@@ -495,6 +495,8 @@ template <bool kDecayToItems>
                   chunk_from_second.offset += first_size;
                 });
 
+  ans.chunks_.back().offset = first_size + second.chunks_.back().offset;
```
tonyastolfi, line 511: Please add regression test for this fix.

src/turtle_kv/core/testing/generate.hpp
```diff
@@ -184,21 +184,29 @@ class RandomResultSetGenerator : public MinMaxSize<usize{1} << 24>
...
     std::vector<EditView> items;
+    
+    for (const KeyView& delete_key : to_delete) {
+      items.emplace_back(delete_key, ValueView::deleted());
+    }
 
     while (items.size() < n) {
       for (usize i = items.size(); i < n; ++i) {
         char ch = '_' + (i & 31);
         items.emplace_back(this->key_generator_(rng, store),
```
tonyastolfi: We should probably check to make sure we don't generate one of the `to_delete` keys, since that would bring the key back to life.

src/turtle_kv/kv_store.cpp
```diff
@@ -677,11 +680,17 @@ StatusOr<ValueView> KVStore::get(const KeyView& key) noexcept /*override*/
         *value = combine(*value, *delta_value);
         if (!value->needs_combine()) {
           this->metrics_.delta_log2_get_count[batt::log2_ceil(observed_deltas_size - i)].add(1);
+          if (value->is_delete()) {
```
tonyastolfi: Consider abstracting to remove the duplicate pattern.

src/turtle_kv/kv_store_scanner.cpp
```diff
@@ -435,15 +435,34 @@ Status KVStoreScanner::set_next_item()
...
+      // represents a deleted item.
+      //
+      if (this->next_item_->value == ValueView::deleted()) {
+        // The terminal value represents a deleted item, so discard it by setting this->next_item_
+        // to None. Then, continue on to the next iteration of the loop, skipping the logic to
+        // advance the current scan_level. We do this because we now need to set the first key
+        // in the current scan_level to this->next_item_ to examine it next.
+        //
+        this->next_item_ = None;
+        continue;
```
tonyastolfi, line 504: Here should be the if needs resume, resume.  And we can keep the flag, just call resume once when going from key to different key.

src/turtle_kv/core/merge_compactor.test.cpp
```diff
@@ -482,4 +487,50 @@ TEST(MergeCompactor, ResultSetDropKeyRange)
   }
 }
 
+TEST(MergeCompactor, ResultSetConcat)
```
tonyastolfi: Good test; I have a few suggestions for improvement:

1. A few comments, to structure the different phases of the test visually, would be nice
2. Have test cases for boundary conditions, like one or the other input to `concat` being empty
3. Verify failure for invalid inputs (like overlapping or swapped key ranges are rejected)
4. It would be good to test that "fragmented" result sets are property concat'ed: so you could drop some key ranges:
    1. drop from first and or second
    2. drop from beginning, middle, or end
5. Have a test where the two inputs are differently-sized

src/turtle_kv/core/testing/generate.hpp
```diff
@@ -184,21 +185,40 @@ class RandomResultSetGenerator : public MinMaxSize<usize{1} << 24>
...
     using Item = typename ResultSet::value_type;
 
     const usize n = this->Super::pick_size(rng);
     std::vector<EditView> items;
 
+    for (const KeyView& delete_key : to_delete) {
+      items.emplace_back(delete_key, ValueView::deleted());
+    }
+    if (items.size() > n) {
+      ResultSet result;
```
tonyastolfi: Should we maybe truncate `items` in this case?  Or just error?

src/turtle_kv/tree/subtree.hpp
```diff
@@ -138,6 +138,19 @@ class Subtree
    */
   StatusOr<Optional<Subtree>> try_split(BatchUpdateContext& context);
 
+  /** \brief Attempts to merge the given Subtree with one of its siblings. If successful, the
+   * newly merged Subtree is returned.
+   *
+   * If no merge, returns None.
+   */
+  StatusOr<Optional<Subtree>> try_merge(BatchUpdateContext& context, Subtree& sibling) noexcept;
```
tonyastolfi: After chatting, new design will be:

1. `sibling` is always the right-sibling
2. `sibling` passed as rvalue
3. modify `this` in place instead of returning a copy on success
4. unify the merge and borrow cases, by having `try_merge` return `None` if its a 2 -> 1 merge, otherwise have it return the new right-sibling if its a 2 -> 2 borrow-style merge; that way the convention is the same as `try_split`.

src/turtle_kv/tree/subtree.hpp
```diff
@@ -138,6 +138,19 @@ class Subtree
...
+
+  /** \brief Attempts to make the Subtree viable by borrowing data from one of its siblings.
```
tonyastolfi: This can be removed from `Subtree` after the previous comment is addressed.

src/turtle_kv/tree/subtree.cpp
```diff
@@ -526,6 +582,51 @@ StatusOr<Optional<Subtree>> Subtree::try_split(BatchUpdateContext& context)
...
+
+      [&](std::unique_ptr<InMemoryLeaf>& leaf) -> StatusOr<Optional<Subtree>> {
+        BATT_CHECK(batt::is_case<std::unique_ptr<InMemoryLeaf>>(sibling.impl_));
+        auto& sibling_leaf_ptr = std::get<std::unique_ptr<InMemoryLeaf>>(sibling.impl_);
+        BATT_CHECK(sibling_leaf_ptr);
+
+        BATT_ASSIGN_OK_RESULT(std::unique_ptr<InMemoryLeaf> merged_leaf,  //
+                              leaf->try_merge(context, std::move(sibling_leaf_ptr)));
+
+        BATT_CHECK_EQ(merged_leaf, nullptr);
```
tonyastolfi: This should be handled the same as lines 622-624 below; then we should have a single generic implementation to replace the two lambdas.

src/turtle_kv/tree/subtree.cpp
```diff
@@ -139,13 +139,10 @@ Status Subtree::apply_batch_update(const TreeOptions& tree_options,
 
           auto new_leaf = std::make_unique<InMemoryLeaf>(llfs::PinnedPage{}, tree_options);
 
-          new_leaf->result_set = update.result_set;
+          new_leaf->result_set = std::move(update.context.decay_batch_to_items(update.result_set));
```
tonyastolfi: no `std::move` is needed here.

src/turtle_kv/tree/subtree.cpp
```diff
@@ -177,7 +174,7 @@ Status Subtree::apply_batch_update(const TreeOptions& tree_options,
 
           BATT_ASSIGN_OK_RESULT(  //
               new_leaf->result_set,
-              update.context.merge_compact_edits(  //
+              update.context.merge_compact_edits</*decay_to_items=*/true>(  //
                   global_max_key(),
                   [&](MergeCompactor& compactor) -> Status {
                     compactor.push_level(update.result_set.live_edit_slices());
```
tonyastolfi: Are we missing a call to `update_has_page_refs` in this case?

src/turtle_kv/tree/subtree.cpp
```diff
@@ -216,7 +213,7 @@ Status Subtree::apply_batch_update(const TreeOptions& tree_options,
```
tonyastolfi, line 201: We should unify this case (most of it) with the Batch + PackedLeaf => InMemoryLeaf case above (generic impl)

src/turtle_kv/tree/subtree.cpp
```diff
@@ -256,6 +253,14 @@ Status Subtree::apply_batch_update(const TreeOptions& tree_options,
           return OkStatus();
         },
         [&](NeedsSplit needs_split) {
+          if (needs_split.too_many_segments && !needs_split.too_many_pivots &&
```
tonyastolfi: Would you please  add a TODO, to revisit this once the latest VLDB changes are merged in?

src/turtle_kv/tree/subtree.cpp
```diff
@@ -306,6 +316,52 @@ Status Subtree::split_and_grow(BatchUpdateContext& context,
...
+
+  return batt::case_of(
+      this->impl_,
+
+      [&](const llfs::PageIdSlot& page_id_slot [[maybe_unused]]) -> Status {
+        return {batt::StatusCode::kUnimplemented};
+      },
+
+      [&](const std::unique_ptr<InMemoryLeaf>& leaf [[maybe_unused]]) -> Status {
+        return OkStatus();
```
tonyastolfi: We should check to see if there are literally no items in the leaf, and if so, replace with an empty tree.

src/turtle_kv/tree/subtree.cpp
```diff
@@ -306,6 +316,52 @@ Status Subtree::split_and_grow(BatchUpdateContext& context,
...
+//
+Status Subtree::flush_and_shrink(BatchUpdateContext& context) noexcept
+{
+  BATT_CHECK(!this->locked_.load());
+
+  return batt::case_of(
+      this->impl_,
+
+      [&](const llfs::PageIdSlot& page_id_slot [[maybe_unused]]) -> Status {
+        return {batt::StatusCode::kUnimplemented};
```
tonyastolfi: Maybe return OkStatus if the page id is invalid?

src/turtle_kv/tree/in_memory_leaf.cpp
```diff
@@ -8,6 +8,74 @@
...
+  buffer.reserve(packed_items.size());
+
+  if (max_tasks == 1) {
+    for (const PackedKeyValue& pkv : packed_items) {
+      buffer.emplace_back(to_edit_view(pkv));
+    }
+  } else {
+    const ParallelAlgoDefaults& algo_defaults = parallel_algo_defaults();
+
+    const auto src_begin = packed_items.begin();
```
tonyastolfi: This code looks good; but it is already available in: batteries/algo/parallel_transform.hpp

src/turtle_kv/tree/in_memory_leaf.cpp
```diff
@@ -149,6 +217,65 @@ auto InMemoryLeaf::make_split_plan() const -> StatusOr<SplitPlan>
   return plan;
 }
 
+//==#==========+==+=+=++=+++++++++++-+-+--+----- --- -- -  -  -   -
+//
+StatusOr<std::unique_ptr<InMemoryLeaf>> InMemoryLeaf::try_merge(
```
tonyastolfi: This looks good for now; please add a TODO to handle the re-balancing case here instead of having re-balance = merge-then-split, as the former is more efficient.

src/turtle_kv/tree/in_memory_leaf.cpp
```diff
@@ -149,6 +217,65 @@ auto InMemoryLeaf::make_split_plan() const -> StatusOr<SplitPlan>
...
+
+  this->set_edit_size_totals(context.compute_running_total(this->result_set));
+
+  return nullptr;
+}
+
+//==#==========+==+=+=++=+++++++++++-+-+--+----- --- -- -  -  -   -
+//
+Status InMemoryLeaf::apply_batch_update(BatchUpdate& update,
+                                        Optional<BoxedSeq<EditSlice>>&& current_result_set) noexcept
```
tonyastolfi: I think we should not pass the second argument here; instead, we can make `this->result_set_` be `Optional`, and lazily initialize it in this function.

src/turtle_kv/tree/subtree.cpp
```diff
@@ -300,6 +307,52 @@ Status Subtree::split_and_grow(BatchUpdateContext& update_context,
...
+{
+  BATT_CHECK(!this->locked_.load());
+
+  BATT_CHECK(!this->is_serialized());
+
+  while (!is_root_viable(this->get_viability())) {
+    // First, try flushing. If flushing makes the root viable, return immediately.
+    //
+    Status flush_status = this->try_flush(context);
+    if (flush_status != OkStatus() && flush_status != batt::StatusCode::kUnavailable) {
```
tonyastolfi: nit: `!flush_status.ok()`

src/turtle_kv/tree/subtree.cpp
```diff
@@ -300,6 +307,47 @@ Status Subtree::split_and_grow(BatchUpdateContext& update_context,
...
 
+//==#==========+==+=+=++=+++++++++++-+-+--+----- --- -- -  -  -   -
+//
+Status Subtree::flush_and_shrink(BatchUpdateContext& context) noexcept
+{
+  BATT_CHECK(!this->locked_.load());
+
+  BATT_CHECK(!this->is_serialized());
+
+  while (!is_root_viable(this->get_viability())) {
```
tonyastolfi: Suggest adding a max retries counter in this loop.  Maybe `kMaxPivots`?

src/turtle_kv/tree/subtree.cpp
```diff
@@ -300,6 +307,47 @@ Status Subtree::split_and_grow(BatchUpdateContext& update_context,
...
+      // Note: At this point, we must have a node and not a leaf, since the `is_root_viable` check
+      // above will return Viable` for a leaf and we break out of the loop in that case.
+      //
+      BATT_REQUIRE_OK(this->try_shrink());
+    }
+  }
+
+  // If the root is a leaf and there are no items in the leaf, set the root to be an empty subtree.
+  //
+  if (batt::is_case<std::unique_ptr<InMemoryLeaf>>(this->impl_)) {
```
tonyastolfi: Looks right, but maybe it should be in `try_shrink` instead..?

src/turtle_kv/tree/subtree.cpp
```diff
@@ -527,6 +575,34 @@ StatusOr<Optional<Subtree>> Subtree::try_split(BatchUpdateContext& context)
       });
 }
 
+//==#==========+==+=+=++=+++++++++++-+-+--+----- --- -- -  -  -   -
+//
+Status Subtree::try_merge(BatchUpdateContext& context, Subtree&& sibling) noexcept
+{
+  BATT_CHECK(!this->locked_.load());
```
tonyastolfi, line 580: Maybe assert/check that this height and sibling height are equal?

src/turtle_kv/tree/in_memory_node.cpp
```diff
@@ -717,6 +738,213 @@ Status InMemoryNode::split_child(BatchUpdateContext& update_context, i32 pivot_i
...
+  if (this->pivot_count() == 1) {
+    return OkStatus();
+  }
+
+  // Decide which sibling to merge with. Edge cases: child that needs merge is the leftmost or
+  // rightmost child in the node.
+  //
+  i32 sibling_i = pivot_i;
+
+  if (pivot_i == 0) {
```
tonyastolfi: Should just be 2 cases.

src/turtle_kv/tree/subtree.cpp
```diff
@@ -527,6 +575,34 @@ StatusOr<Optional<Subtree>> Subtree::try_split(BatchUpdateContext& context)
...
+      [&](const llfs::PageIdSlot& page_id_slot) -> Status {
+        BATT_PANIC() << "Cannot try merging a serialized subtree!";
+
+        return {batt::StatusCode::kUnimplemented};
+      },
+
+      [&](auto& in_memory) -> Status {
+        using PtrT = std::decay_t<decltype(in_memory)>;
+
+        BATT_CHECK(batt::is_case<PtrT>(sibling.impl_));
```
tonyastolfi: Would be good to add a message that interprets what this panic means.

src/turtle_kv/tree/subtree.cpp
```diff
@@ -527,6 +575,34 @@ StatusOr<Optional<Subtree>> Subtree::try_split(BatchUpdateContext& context)
...
+        auto& sibling_ptr = std::get<PtrT>(sibling.impl_);
+        BATT_CHECK(sibling_ptr);
```
tonyastolfi: nit: `BATT_CHECK_NOT_NULLPTR`

src/turtle_kv/tree/in_memory_node.cpp
```diff
@@ -717,6 +738,213 @@ Status InMemoryNode::split_child(BatchUpdateContext& update_context, i32 pivot_i
...
+  //
+  for (Level& level : this->update_buffer.levels) {
+    if (batt::is_case<SegmentedLevel>(level)) {
+      SegmentedLevel& segmented_level = std::get<SegmentedLevel>(level);
+      in_segmented_level(*this,
+                         segmented_level,
+                         update_context.page_loader,
+                         update_context.overcommit)
+          .merge_pivots(left_pivot_i, right_pivot_i);
+    }
```
tonyastolfi: Maybe comment why we don't care about the other cases?

Should probably be a `batt::case_of`

src/turtle_kv/tree/in_memory_node.cpp
```diff
@@ -717,6 +738,213 @@ Status InMemoryNode::split_child(BatchUpdateContext& update_context, i32 pivot_i
...
+  //
+  this->children.erase(this->children.begin() + right_pivot_i);
+
+  // Update pending_bytes. The leftmost of {subtree, sibling} should be incremented by the removed
+  // subtree's pending bytes values. Erase the pending bytes of the removed subtree.
+  //
+  this->pending_bytes[left_pivot_i] += this->pending_bytes[right_pivot_i];
+  this->pending_bytes.erase(this->pending_bytes.begin() + right_pivot_i);
+
+  bool is_pending_bytes_exact = get_bit(this->pending_bytes_is_exact, left_pivot_i) &
```
tonyastolfi: `&`  -> `&&`

src/turtle_kv/tree/in_memory_leaf.cpp
```diff
@@ -150,6 +195,38 @@ auto InMemoryLeaf::make_split_plan() const -> StatusOr<SplitPlan>
...
+//==#==========+==+=+=++=+++++++++++-+-+--+----- --- -- -  -  -   -
+//
+Status InMemoryLeaf::try_merge(BatchUpdateContext& context,
+                               std::unique_ptr<InMemoryLeaf> sibling) noexcept
+{
+  BATT_CHECK(this->result_set);
+  BATT_CHECK(sibling->result_set);
+
+  if (sibling->result_set->empty()) {
+    BATT_CHECK(batt::is_case<Viable>(this->get_viability()));
```
tonyastolfi: Message would be helpful here.

src/turtle_kv/tree/in_memory_leaf.cpp
```diff
@@ -150,6 +195,38 @@ auto InMemoryLeaf::make_split_plan() const -> StatusOr<SplitPlan>
...
+
+  if (this->result_set->empty()) {
+    BATT_CHECK(batt::is_case<Viable>(sibling->get_viability()));
+    this->pinned_leaf_page_ = std::move(sibling->pinned_leaf_page_);
+    this->result_set = std::move(sibling->result_set);
+    this->shared_edit_size_totals_ = sibling->shared_edit_size_totals_;
+    this->edit_size_totals = std::move(sibling->edit_size_totals);
+    return OkStatus();
+  }
+
```
tonyastolfi, line 224: We should probably retain a pin on the sibling's leaf page in `this`; that means we may need to change `InMemoryLeaf::pinned_leaf_page_` into `pinned_leaf_pages_`.

src/turtle_kv/tree/in_memory_leaf.cpp
vidyasilai: Done

src/turtle_kv/tree/in_memory_node.cpp
vidyasilai: Done

src/turtle_kv/tree/in_memory_node.cpp
vidyasilai: Done

src/turtle_kv/tree/subtree.cpp
vidyasilai: Done

src/turtle_kv/tree/subtree.cpp
vidyasilai: Done

src/turtle_kv/tree/in_memory_node.cpp
vidyasilai: Done

src/turtle_kv/tree/subtree.cpp
vidyasilai: Done

src/turtle_kv/tree/subtree.cpp
vidyasilai: Done

src/turtle_kv/tree/subtree.cpp
vidyasilai: Done

src/turtle_kv/tree/subtree.cpp
vidyasilai: Done

src/turtle_kv/tree/in_memory_leaf.cpp
vidyasilai: In regard to turning `pinned_leaf_page_` into a vector of pinned pages, what are your thoughts on keeping `pinned_leaf_page_` as is and adding another member variable to `InMemoryLeaf` that is of type `std::vector<PinnedPage>` that holds pins on sibling pages as a result of merges? My reason for suggesting this is due to the logic in `InMemoryLeaf::apply_batch_update`, where we check the state of `pinned_leaf_page_` to determine how to extract `result_set`.

src/turtle_kv/tree/subtree.cpp
```diff
@@ -300,6 +307,43 @@ Status Subtree::split_and_grow(BatchUpdateContext& update_context,
...
+      return flush_status;
+    }
+
+    // Nothing was available to flush since either the node's update buffer is empty or we have
+    // a leaf root. If possible and necessary, try collapsing one level of the tree.
+    //
+    if (flush_status == batt::StatusCode::kUnavailable) {
+      BATT_REQUIRE_OK(this->try_shrink());
+    }
+
```
tonyastolfi, line 338: After a successful flush, we should do something here to see if the root object has changed; if it has, then we should reset the `retries` counter, since limiting to `kMaxPivots` is based on the assumption of a single node.

src/turtle_kv/tree/subtree_viability.hpp
```diff
@@ -160,4 +160,21 @@ inline bool normal_flush_might_fix_root(const SubtreeViability& viability)
...
+  return batt::case_of(
+      viability,
+      [](const Viable&) {
+        return true;
+      },
+      [](const NeedsSplit&) {
+        return false;
+      },
+      [](const NeedsMerge& needs_merge) {
+        return !needs_merge.single_pivot;
```
tonyastolfi: `&& !needs_merge.no_items`

src/turtle_kv/tree/packed_leaf_page.hpp
```diff
@@ -441,6 +441,9 @@ inline usize PackedLeafLayoutPlan::compute_trie_step_size() const
     step_size = 1;
   } else if (this->key_count < 256) {
     step_size = this->key_count / 16;
```
tonyastolfi: Consider rounding up instead..?

src/turtle_kv/tree/active_pivots_set.hpp
```diff
@@ -173,6 +178,33 @@ class ActivePivotsSet128 : public ActivePivotsSetBase<std::array<u64, 2>>
       this->bit_set_[1] = 0;
     }
   }
+
+  /** \brief Removes the specified number (`count`) pivots from the end of the bit set.
+   */
+  BATT_ALWAYS_INLINE void pop_back_pivots(i32 count)
```
tonyastolfi: maybe `push_front_pivots`?

src/turtle_kv/tree/in_memory_node.cpp
```diff
@@ -1695,6 +1995,390 @@ void InMemoryNode::UpdateBuffer::SegmentedLevel::check_items_sorted(
...
+          segment.active_pivots.pop_back_pivots(node_pivot_count);
+        };
+
+        this->segments.insert(this->segments.end(),
+                              std::make_move_iterator(right_segmented_level.segments.begin()),
+                              std::make_move_iterator(right_segmented_level.segments.end()));
+
+        // Erase potential duplicate segments that have resulted from a previous split.
+        //
+        this->segments.erase(std::unique(this->segments.begin(),
```
tonyastolfi: We do need to do this de-duplication step, but we need to calculate:

1. union of the active pivots set
2. "joining" of the portions of the piecewise-filter that correspond to each sibling's "side."

src/turtle_kv/tree/algo/segments.hpp
```diff
@@ -141,6 +141,18 @@ struct SegmentAlgorithms {
...
   }
 
+  /** \brief Merges the two given pivots, effectively erasing `right_pivot`.
+   */
+  template <typename LevelT>
+  [[nodiscard]] void merge_pivots(i32 left_pivot, i32 right_pivot, const LevelT& level)
+  {
+    bool left_is_active = this->segment_.is_pivot_active(left_pivot);
+    bool right_is_active = this->segment_.is_pivot_active(right_pivot);
+    this->segment_.set_pivot_active(left_pivot, left_is_active | right_is_active);
```
tonyastolfi: `|` -> `||`

src/turtle_kv/tree/in_memory_node.hpp
```diff
@@ -595,10 +754,25 @@ struct InMemoryNode {
    */
   Status try_flush(BatchUpdateContext& context);
 
+  /** \brief Attempt to collapse one level of the tree. Returns the node's single pivot.
+   */
+  Subtree try_shrink();
```
tonyastolfi: Either:

`Optional<Subtree> try_shrink()` / `StatusOr<Subtree> try_shrink()`

Or:

`Subtree shrink()` / `Subtree shrink_or_panic()`

src/turtle_kv/tree/in_memory_node.cpp
```diff
@@ -717,6 +731,170 @@ Status InMemoryNode::split_child(BatchUpdateContext& update_context, i32 pivot_i
...
 }
 
+//==#==========+==+=+=++=+++++++++++-+-+--+----- --- -- -  -  -   -
+//
+Subtree InMemoryNode::try_shrink()
+{
+  BATT_CHECK_EQ(this->children.size(), 1);
+  BATT_CHECK_EQ(this->pending_bytes.size(), 1);
+  BATT_CHECK_EQ(this->pending_bytes[0], 0);
+
```
tonyastolfi, line 742: Maybe also sanity check that there are no non-empty levels?

src/turtle_kv/tree/in_memory_node.cpp
```diff
@@ -717,6 +731,170 @@ Status InMemoryNode::split_child(BatchUpdateContext& update_context, i32 pivot_i
   return OkStatus();
 }
 
+//==#==========+==+=+=++=+++++++++++-+-+--+----- --- -- -  -  -   -
+//
+Subtree InMemoryNode::try_shrink()
+{
+  BATT_CHECK_EQ(this->children.size(), 1);
+  BATT_CHECK_EQ(this->pending_bytes.size(), 1);
```
tonyastolfi, line 739: Consider allowing shrink with non-empty levels (pending bytes > 0), when all the levels of this node and its child will fit in a single buffer.

src/turtle_kv/tree/in_memory_node.cpp
```diff
@@ -953,10 +1127,19 @@ usize InMemoryNode::total_segment_filter_cut_points() const
...
         [](const SegmentedLevel& segmented_level) -> usize {
+          return segmented_level.segment_filter_cut_points();
+        },
+        [](const HybridLevel& hybrid_level) -> usize {
           usize n = 0;
-          for (const Segment& segment : segmented_level.segments) {
-            n += segment.num_cut_points();
+
+          auto sub_levels = hybrid_level.get_levels();
+          for (auto iter = sub_levels.begin(); iter != sub_levels.end(); ++iter) {
```
tonyastolfi: nit: for each loop

src/turtle_kv/tree/in_memory_node.cpp
```diff
@@ -1327,6 +1510,31 @@ StatusOr<ValueView> InMemoryNode::find_key_in_level(usize level_i,
       [&](const SegmentedLevel& segmented_level) -> StatusOr<ValueView> {
         return in_segmented_level(*this, segmented_level, *query.page_loader, query.overcommit())
             .find_key(key_pivot_i, query);
+      },
+      [&](const HybridLevel& hybrid_level) -> StatusOr<ValueView> {
+        StatusOr<ValueView> result{Status{batt::StatusCode::kNotFound}};
+
+        auto sub_levels = hybrid_level.get_levels();
+        for (auto iter = sub_levels.begin(); iter != sub_levels.end(); ++iter) {
```
tonyastolfi: foreach

src/turtle_kv/tree/in_memory_node.cpp
```diff
@@ -1327,6 +1510,31 @@ StatusOr<ValueView> InMemoryNode::find_key_in_level(usize level_i,
...
+              },
+              [&](const SegmentedLevel& segmented_sub_level) -> StatusOr<ValueView> {
+                return in_segmented_level(*this,
+                                          segmented_sub_level,
+                                          *query.page_loader,
+                                          query.overcommit())
+                    .find_key(key_pivot_i, query);
+              });
+
+          if (result.ok()) {
```
tonyastolfi: maybe `if (result.ok() || result != batt::StatusCode::kNotFound) {...`?

src/turtle_kv/tree/in_memory_node.cpp
tonyastolfi: It seems like too much work to iterate over all the "sub-levels."  But I suppose we don't expect there to be too many of them, so it's probably fine.  Maybe it just deserves a TODO, opportunity for optimization in the future?

src/turtle_kv/tree/in_memory_node.cpp
tonyastolfi: Let's pull the `find_key` code for each level type into that type, so we can share this code (and the code in the SegmentedLevel case) below in HybridLevel.

src/turtle_kv/tree/in_memory_node.cpp
```diff
@@ -1480,6 +1701,38 @@ StatusOr<llfs::PinnedPage> Segment::load_leaf_page(llfs::PageLoader& page_loader
...
+{
+  return batt::case_of(
+      sibling_level,
+      [&](EmptyLevel&) -> Level {
+        return *this;
+      },
+      [&](MergedLevel& right_merged_level) -> Level {
+        return this->concat(std::move(right_merged_level));
+      },
+      [&](SegmentedLevel& right_segmented_level) -> Level {
```
tonyastolfi: Consider combining this lambda and the next one, using a generic `auto&` param and a `requires` clause just to make sure we *only* accept `SegmentedLevel` and `HybridLevel` cases.

src/turtle_kv/tree/in_memory_node.cpp
```diff
@@ -1694,6 +1947,413 @@ void InMemoryNode::UpdateBuffer::SegmentedLevel::check_items_sorted(
...
+}
+
+//==#==========+==+=+=++=+++++++++++-+-+--+----- --- -- -  -  -   -
+//
+bool InMemoryNode::UpdateBuffer::SegmentedLevel::set_pivot_items_flushed(
+    const InMemoryNode& node,
+    BatchUpdateContext& update_context,
+    usize pivot_i,
+    const CInterval<KeyView>& flush_key_crange,
+    Status segment_load_status)
```
tonyastolfi: Probably should be passed by pointer or ref?  The Update in here won't have any side-effects visible outside.

src/turtle_kv/tree/in_memory_node.cpp
```diff
@@ -1694,6 +1947,413 @@ void InMemoryNode::UpdateBuffer::SegmentedLevel::check_items_sorted(
...
+}
+
+//==#==========+==+=+=++=+++++++++++-+-+--+----- --- -- -  -  -   -
+//
+Level InMemoryNode::UpdateBuffer::HybridLevel::merge(Level&& sibling_level, usize node_pivot_count)
+{
+  batt::case_of(
+      sibling_level,
+      [&](EmptyLevel&) {
+
```
tonyastolfi: nit: blank line

src/turtle_kv/tree/in_memory_node.cpp
```diff
@@ -1694,6 +1947,413 @@ void InMemoryNode::UpdateBuffer::SegmentedLevel::check_items_sorted(
...
+      },
+      [&](MergedLevel& right_merged_level) {
+        this->add_new_sub_level(std::move(right_merged_level));
+      },
+      [&](SegmentedLevel& right_segmented_level) {
+        right_segmented_level.push_front_pivots(node_pivot_count);
+        this->add_new_sub_level(std::move(right_segmented_level));
```
tonyastolfi: If we always call `push_front_pivots` right before `add_new_sub_level`, consider doing it *inside* the `add_new_sub_level` function so we can't forget that step.

src/turtle_kv/tree/in_memory_node.cpp
```diff
@@ -1694,6 +1947,413 @@ void InMemoryNode::UpdateBuffer::SegmentedLevel::check_items_sorted(
...
+    TreeSerializeContext& context)
+{
+  usize total_segment_count = 0;
+
+  for (auto iter = this->levels.begin(); iter != this->levels.end(); ++iter) {
+    if (batt::is_case<MergedLevel>(*iter)) {
+      MergedLevel& merged_sub_level = std::get<MergedLevel>(*iter);
+      BATT_ASSIGN_OK_RESULT(usize sub_level_total, merged_sub_level.start_serialize(node, context));
+      total_segment_count += sub_level_total;
+    }
```
tonyastolfi: We also need to update `total_segment_count` for the SegmentedLevel sub-level case.

src/turtle_kv/tree/in_memory_node.cpp
```diff
@@ -1694,6 +1947,413 @@ void InMemoryNode::UpdateBuffer::SegmentedLevel::check_items_sorted(
...
+          *iter = *new_segmented_sub_level;
+          r_metrics.serialized_buffer_segment_count.add(new_segmented_sub_level->segment_count());
+
+          return OkStatus();
+        },
+        [&](SegmentedLevel& segmented_sub_level) {
+          r_metrics.serialized_buffer_segment_count.add(segmented_sub_level.segment_count());
+          return OkStatus();
+        }));
+  }
```
tonyastolfi: +1 on your in-person comment to combine these two loops (avoiding the awkwardness of updating the variant in `*iter` while inside the `case_of`.

tonyastolfi:
Finished reading the HybridLevel stuff -- mostly good, some opportunities for DRY.

We had finished reviewing the Subtree stuff earlier, so that's ok.

TODO: 
  - `InMemoryNode::try_merge`
  - `InMemoryNode::merge_child`
  - Test cases
  - (Tony, offline) review open threads that Vidya replied to.

src/turtle_kv/tree/subtree.cpp
vidyasilai: Done

src/turtle_kv/tree/subtree_viability.hpp
vidyasilai: Done

src/turtle_kv/tree/packed_leaf_page.hpp
vidyasilai: Done

src/turtle_kv/tree/active_pivots_set.hpp
vidyasilai: Done

src/turtle_kv/tree/algo/segments.hpp
vidyasilai: Done

src/turtle_kv/tree/in_memory_node.hpp
vidyasilai: Fixed to `shrink_or_panic`

src/turtle_kv/tree/in_memory_node.cpp
vidyasilai: Done

src/turtle_kv/tree/in_memory_node.cpp
vidyasilai: Added a TODO in subtree.cpp for this.

src/turtle_kv/tree/in_memory_node.cpp
vidyasilai: Fixed

src/turtle_kv/tree/in_memory_node.cpp
vidyasilai: Fixed

src/turtle_kv/tree/in_memory_node.cpp
vidyasilai: Fixed

src/turtle_kv/tree/in_memory_node.cpp
vidyasilai: Added a TODO.

src/turtle_kv/tree/in_memory_node.cpp
vidyasilai: Done

src/turtle_kv/tree/in_memory_node.cpp
vidyasilai: Done

src/turtle_kv/tree/in_memory_node.cpp
vidyasilai: Fixed

src/turtle_kv/tree/in_memory_node.cpp
vidyasilai: Fixed

src/turtle_kv/tree/in_memory_node.cpp
vidyasilai: Done

src/turtle_kv/tree/in_memory_node.cpp
vidyasilai: Done

src/turtle_kv/tree/in_memory_node.cpp
vidyasilai: Done

src/turtle_kv/tree/in_memory_leaf.cpp
tonyastolfi: Yeah, that makes sense.

src/turtle_kv/tree/in_memory_node.cpp
```diff
@@ -1694,6 +1918,492 @@ void InMemoryNode::UpdateBuffer::SegmentedLevel::check_items_sorted(
...
+
+    StatusOr<SegmentedLevel> new_segmented_sub_level = batt::case_of(
+        sub_level,
+        [&](MergedLevel& merged_sub_level) -> StatusOr<SegmentedLevel> {
+          return merged_sub_level.finish_serialize(node, context);
+        },
+        [&](SegmentedLevel& segmented_sub_level) -> StatusOr<SegmentedLevel> {
+          r_metrics.serialized_buffer_segment_count.add(segmented_sub_level.segment_count());
+
+          final_segmented_level.segments.insert(
```
tonyastolfi: Remove (reminder)

src/turtle_kv/tree/in_memory_node.cpp
```diff
@@ -1694,6 +1918,492 @@ void InMemoryNode::UpdateBuffer::SegmentedLevel::check_items_sorted(
...
+{
+  batt::case_of(
+      sibling_level,
+      [&](EmptyLevel&) {
+      },
+      [&](MergedLevel& right_merged_level) {
+        this->add_new_sub_level(std::move(right_merged_level));
+      },
+      [&](SegmentedLevel& right_segmented_level) {
+        right_segmented_level.push_front_pivots(node_pivot_count);
```
tonyastolfi: Reminder: remove this line

src/turtle_kv/tree/in_memory_node.cpp
```diff
@@ -1694,6 +1918,492 @@ void InMemoryNode::UpdateBuffer::SegmentedLevel::check_items_sorted(
...
+        this->add_new_sub_level(std::move(right_segmented_level));
+      },
+      [&](HybridLevel& right_hybrid_level) {
+        right_hybrid_level.push_front_pivots(node_pivot_count);
```
tonyastolfi: Reminder: remove this line

src/turtle_kv/tree/in_memory_node.cpp
```diff
@@ -1694,6 +1918,492 @@ void InMemoryNode::UpdateBuffer::SegmentedLevel::check_items_sorted(
...
+
+        return new_hybrid_level;
+      },
+      [&](SegmentedLevel& right_segmented_level) -> Level {
+        // First shift the right level's bit set to the left by the number of pivots
+        // in the left node.
+        //
+        right_segmented_level.push_front_pivots(node_pivot_count);
+
+        this->segments.insert(this->segments.end(),
```
tonyastolfi: Let's handle the leaf page segment de-dup case only at the seam.

src/turtle_kv/tree/in_memory_node.cpp
```diff
@@ -1694,6 +1918,492 @@ void InMemoryNode::UpdateBuffer::SegmentedLevel::check_items_sorted(
...
+                                                                         i32 key_pivot_i) const
+{
+  return in_segmented_level(node, *this, *query.page_loader, query.overcommit())
+      .find_key(key_pivot_i, query);
+}
+
+//==#==========+==+=+=++=+++++++++++-+-+--+----- --- -- -  -  -   -
+//
+Level InMemoryNode::UpdateBuffer::SegmentedLevel::merge(Level&& sibling_level,
+                                                        usize node_pivot_count)
```
tonyastolfi: Since this function is (legitimately) capable of returning `*this`, it might be helpful to qualify it with the `&&` in its declaration.

src/turtle_kv/tree/in_memory_node.hpp
```diff
@@ -74,7 +82,12 @@ struct InMemoryNode {
   struct UpdateBuffer {
     using Self = UpdateBuffer;
 
+    struct EmptyLevel;
```
tonyastolfi: For each of these, let's give them their own header.

The best way I've found to do this is to give the classes a longer name, and place them all in the top-level `turtle_kv` namespace.  For example:

```c++
namespace turtle_kv {
class InMemoryNodeEmptyLevel;
class InMemoryNodeMergedLevel;
// etc.
}
```

and then type-alias them back into the `InMemoryNode` class.

src/turtle_kv/tree/in_memory_node.cpp
vidyasilai: Done

src/turtle_kv/tree/in_memory_node.cpp
vidyasilai: Done

src/turtle_kv/tree/in_memory_node.cpp
vidyasilai: Done

src/turtle_kv/tree/in_memory_node.cpp
vidyasilai: Done, also added for `HybridLevel`.

src/turtle_kv/tree/in_memory_node.cpp
vidyasilai: Done

src/turtle_kv/tree/in_memory_node.hpp
vidyasilai: Done

src/turtle_kv/tree/in_memory_node_hybrid_level.hpp
```diff
@@ -0,0 +1,179 @@
...
+          std::get<InMemoryNodeSegmentedLevel>(*this->back());
+      BATT_CHECK_NOT_NULLPTR(left_sub_level.back());
+
+      left_sub_level.deduplicate(right_level);
+    }
+
+    this->add_new_sub_level(std::move(right_level));
+  }
+
+  void deduplicate_and_add_sub_level(InMemoryNodeHybridLevel&& right_level,
```
tonyastolfi: I think the repeated code between this overload and the SegmentedLevel one just before can largely be eliminated if we use `if constexpr`.  Check out the example here: https://en.cppreference.com/cpp/utility/variant/visit2

src/turtle_kv/tree/in_memory_node_segmented_level.hpp
```diff
@@ -0,0 +1,319 @@
...
+    //
+    bool first_element_live = !live_ranges.empty() && live_ranges.front().lower_bound ==
+                                                          PiecewiseFilter<u32>::kMinLowerBound;
+    bool ends_at_max = !live_ranges.empty() &&
+                       live_ranges.back().upper_bound == PiecewiseFilter<u32>::kMaxUpperBound;
+
+    return live_ranges.size() * 2 - (first_element_live ? 1 : 0) - (ends_at_max ? 1 : 0);
+  }
+
+  void deduplicate(const InMemoryNodeSegment& other)
```
tonyastolfi: I feel like it may be more robust to move the check for leaf page id being the same into this function; we can have this function return `[[nodiscard]] bool` to indicate whether the de-duplication succeeded.  The `[[nodiscard]]` forces the caller to respond, by either discarding the right-hand segment or taking it.

src/turtle_kv/tree/in_memory_node_segmented_level.cpp
```diff
@@ -0,0 +1,402 @@
...
+  BATT_CHECK_NOT_NULLPTR(right_level.front());
+  if (this->back()->get_leaf_page_id() == right_level.front()->get_leaf_page_id()) {
+    right_level.front()->deduplicate(*this->back());
+    this->segments.pop_back();
+  }
+}
+
+//==#==========+==+=+=++=+++++++++++-+-+--+----- --- -- -  -  -   -
+//
+void InMemoryNodeSegmentedLevel::deduplicate(InMemoryNodeHybridLevel& right_level,
```
tonyastolfi: This function is so similar to its overload just above... I think here too, `if constexpr` will allow us to cleanly write both as a single function.

src/turtle_kv/tree/in_memory_node_hybrid_level.hpp
vidyasilai: Done

src/turtle_kv/tree/in_memory_node_segmented_level.hpp
vidyasilai: Done

src/turtle_kv/tree/in_memory_node_segmented_level.cpp
vidyasilai: Done

src/turtle_kv/tree/in_memory_node_hybrid_level.cpp
```diff
@@ -0,0 +1,435 @@
...
+  right_level.push_front_pivots(push_pivot_count);
+
+  BATT_CHECK_NOT_NULLPTR(this->back());
+  BATT_CHECK_NOT_NULLPTR(right_level.front());
+
+  // Attempt deduplication if the last sub level in this HybridLevel is a SegmentedLevel.
+  //
+  bool attempt_deduplication = batt::is_case<SegmentedLevel>(*this->back());
+
+  if constexpr (std::is_same_v<std::decay_t<T>, HybridLevel>) {
```
tonyastolfi, line 317: +1: Excellent!  This perfectly captures the difference and commonality between the two type cases.

PR #6, opened 2026-01-09
KVStore::remove Implementation Part 1

vidyasilai:
First PR for the implementation of `KVStore::remove`.

This PR includes the following changes:

- Decay to items support
- Client facing function updates in the `KVStore` and `KVStoreScanner` classes
- Test infrastructure updates and refactoring (to be continued in the next PR, since more test cases are incoming...)
- Bug fix (and corresponding test cases) for `MergeCompactor::concat`

### DIFF
--- a/src/turtle_kv/core/merge_compactor.test.cpp
+++ b/src/turtle_kv/core/merge_compactor.test.cpp
@@ -5,6 +5,7 @@
 #include <gmock/gmock.h>
 #include <gtest/gtest.h>
 
+#include <turtle_kv/core/testing/generate.hpp>
 #include <turtle_kv/import/env.hpp>
 
 #include <batteries/stream_util.hpp>
@@ -23,7 +24,10 @@ using namespace batt::int_types;
 using batt::as_seq;
 using batt::WorkerPool;
 
+using llfs::StableStringStore;
+
 using turtle_kv::CInterval;
+using turtle_kv::DecayToItem;
 using turtle_kv::EditSlice;
 using turtle_kv::EditView;
 using turtle_kv::getenv_as;
@@ -38,6 +42,8 @@ using turtle_kv::Slice;
 using turtle_kv::Status;
 using turtle_kv::StatusOr;
 using turtle_kv::ValueView;
+
+using turtle_kv::testing::RandomStringGenerator;
 
 namespace seq = turtle_kv::seq;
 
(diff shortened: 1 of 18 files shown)


src/turtle_kv/core/merge_compactor.test.cpp
```diff
@@ -482,4 +488,176 @@ TEST(MergeCompactor, ResultSetDropKeyRange)
...
+      }
+      keys_set.emplace(key);
+      this->all_edits_.emplace_back(key,
+                                    ValueView::from_str(this->store_.store(std::string(100, 'a'))));
+    }
+
+    if (needs_sort) {
+      std::sort(this->all_edits_.begin(), this->all_edits_.end(), KeyOrder{});
+    } else {
+      if (std::is_sorted(this->all_edits_.begin(), this->all_edits_.end(), KeyOrder{})) {
```
tonyastolfi: I think this belongs outside this function, as it is IMO hard to understand why `generate_edits` should have such a behavior without the context of the test point that intentionally hands bad input to concat to trigger a panic (which is a good test, btw).

Also, I don't think this swap is accomplishing what we need it to in that test, as `concat` doesn't actually test for overall sortedness but just checks the last item of the first input against the first item of the second; in fact, this swap doesn't guarantee that there is an incorrect ordering in the _middle_ of the generated edits, which is what we need.  So, in the test that calls `generate_edits` (currently with `needs_sort=false`), we should swap `first.back()` with `second.front()` _after_ splitting the generated edits..

I don't have a problem with `generate_edits` sorting its output in all cases; it isn't wasted work because it is good for the death test to create a scenario where the input to `concat` is correct except for the fact that those two items are swapped.

src/turtle_kv/core/merge_compactor.test.cpp
```diff
@@ -482,4 +488,176 @@ TEST(MergeCompactor, ResultSetDropKeyRange)
...
+
+  MergeCompactor::ResultSet<false> concatenated_result_set =
+      this->concat(std::move(first), std::move(second), DecayToItem<false>{});
+
+  // Concatenated ResultSet should have the same size as the original edit batch, and should
+  // also contain the same items in the same order.
+  //
+  this->verify_result_set(concatenated_result_set, this->all_edits_);
+
+  // Now, repeat the process qith unequal sized inputs
```
tonyastolfi: nit: typo: `qith` -> `with`

src/turtle_kv/core/merge_compactor.test.cpp
```diff
@@ -482,4 +488,176 @@ TEST(MergeCompactor, ResultSetDropKeyRange)
...
+  // Finally, test with empty input.
+  //
+  first = {};
+  second.assign(this->all_edits_.begin(), this->all_edits_.begin() + (n / 4));
+
+  concatenated_result_set = this->concat(std::move(first), std::move(second), DecayToItem<false>{});
+
+  this->verify_result_set(concatenated_result_set,
+                          {this->all_edits_.begin(), this->all_edits_.begin() + (n / 4)});
+
```
tonyastolfi, line 603: How about a test case with the first non-empty and the second empty?

src/turtle_kv/tree/in_memory_leaf.cpp
```diff
@@ -149,12 +150,54 @@ auto InMemoryLeaf::make_split_plan() const -> StatusOr<SplitPlan>
...
+  // applying updates to.
+  //
+  BATT_CHECK_IMPLIES(!current_result_set, !this->pinned_leaf_page_ && !this->result_set);
+
+  BATT_ASSIGN_OK_RESULT(this->result_set,
+                        update.context.merge_compact_edits</*decay_to_items=*/true>(
+                            global_max_key(),
+                            [&](MergeCompactor& compactor) -> Status {
+                              compactor.push_level(update.result_set.live_edit_slices());
+                              if (current_result_set) {
```
tonyastolfi: What does it mean if `current_result_set` is false here (from a "big-picture" perspective)?  Is this a case that should ever occur?  It seems to me that leaves without any items in them should be elided before we ever get into this code, but maybe I'm missing/forgetting something..?

src/turtle_kv/tree/in_memory_leaf.cpp
```diff
@@ -149,12 +150,54 @@ auto InMemoryLeaf::make_split_plan() const -> StatusOr<SplitPlan>
   return plan;
 }
 
+//==#==========+==+=+=++=+++++++++++-+-+--+----- --- -- -  -  -   -
+//
+Status InMemoryLeaf::apply_batch_update(BatchUpdate& update) noexcept
+{
+  Optional<BoxedSeq<EditSlice>> current_result_set = None;
```
tonyastolfi: nit: this isn't actually a `ResultSet`; suggest renaming to something like `current_edits`, `current_edit_seq`, or similar.

src/turtle_kv/tree/subtree_viability.hpp
```diff
@@ -107,5 +107,4 @@ inline bool compacting_levels_might_fix(const SubtreeViability& viability)
                !needs_split.too_many_pivots;
       });
 }
-
```
tonyastolfi, line 110: nit: revert

src/turtle_kv/tree/testing/random_leaf_generator.hpp
```diff
@@ -59,7 +59,7 @@ class RandomLeafGenerator
 
     // Generate a sorted run of random key/value pairs.
     //
-    result.result_set = this->items_generator_(decay_to_items, rng, store);
+    result.result_set = this->items_generator_(decay_to_items, rng, store, {});
```
tonyastolfi: Please change:

```c++
result.result_set = this->items_generator_(decay_to_items, rng, store, /*deleted=*/{});
```

tonyastolfi:
Thanks for this smaller MR, Vidya!  Left a few comments, should be pretty quick to address.

src/turtle_kv/tree/in_memory_leaf.cpp
vidyasilai: This is the case where we have an empty tree. In this case, we won't enter either of the branches above since there is no initialized `ResultSet` or valid `PinnedPage` for the leaf, so `current_result_set` remains as `None`.

src/turtle_kv/core/merge_compactor.test.cpp
vidyasilai: Changed `generate_edits` such that it is always sorting the edits, and performed the swap only in the bad input test case.

src/turtle_kv/core/merge_compactor.test.cpp
vidyasilai: Fixed

src/turtle_kv/core/merge_compactor.test.cpp
vidyasilai: Added this test.

src/turtle_kv/tree/in_memory_leaf.cpp
vidyasilai: Renamed to `current_edits`.

src/turtle_kv/tree/subtree_viability.hpp
vidyasilai: Fixed

src/turtle_kv/tree/testing/random_leaf_generator.hpp
vidyasilai: Fixed

src/turtle_kv/tree/in_memory_leaf.cpp
tonyastolfi: Right.  Thanks!

PR #8, opened 2026-01-23
Add support for page cache overcommit

tonyastolfi:
### `TURTLE_KV_PROFILE_QUERIES`/`TURTLE_KV_PROFILE_QUERIES`

Added to `turtle_kv/config.hpp` to control whether metric collection is compiled along critical paths.

### `BatchUpdateMetrics`

Metrics are now collected for batch updates in particular; these are propagated through the call stack via the BatchUpdate struct; this motivates changing a lot of function signatures that previously received PageLoader, to now take a BatchUpdate ref.  This also allowed some cleanup inside the node algorithms code to update pending bytes counts.

### `try_split`/`try_split_direct`

Refacted InMemoryNode splitting to better separate the core "direct" splitting code from the outer retry loop that detects certain exceptional conditions and applies heuristic fixes.

### Trie Index building (`PackedLeafPage`)

Added some new code to handle edge cases (e.g. leaf contains a single key, item count is less than 16, etc.)

### DIFF
--- a/src/turtle_kv/kv_store.cpp
+++ b/src/turtle_kv/kv_store.cpp
@@ -2,6 +2,7 @@
 //
 
 #include <turtle_kv/config.hpp>
+#include <turtle_kv/on_page_cache_overcommit.hpp>
 
 #include <turtle_kv/checkpoint_log.hpp>
 #include <turtle_kv/file_utils.hpp>
@@ -818,6 +819,9 @@ Status KVStore::update_checkpoint(const State* observed_state)
   });
 
   for (;;) {
+    // Multiple threads are potentially racing to install a new active MemTable;
+    // the important thing is that the old mem table has been replaced.
+    //
     if (observed_state->mem_table_ != old_mem_table) {
       BATT_CHECK_NE(observed_state, new_state);
       BATT_CHECK_EQ(new_state->use_count(), 1);
@@ -994,6 +998,7 @@ std::unique_ptr<DeltaBatch> KVStore::compact_memtable(boost::intrusive_ptr<MemTa
   TURTLE_KV_COLLECT_LATENCY(this->metrics_.compact_batch_latency,
                             delta_batch->merge_compact_edits());
 
+  this->metrics_.batch_count.add(1);
   this->metrics_.batch_edits_count.add(delta_batch->result_set_size());
 
   return delta_batch;
@@ -1032,14 +1037,21 @@ void KVStore::checkpoint_update_thread_main()
 StatusOr<std::unique_ptr<CheckpointJob>> KVStore::apply_batch_to_checkpoint(
     std::unique_ptr<DeltaBatch>&& delta_batch)
 {
+  // Allow the page cache to be temporarily overcommitted so we don't deadlock; but if overcommit is
+  // triggered, we immediately truncate the checkpoint to try to unpin pages ASAP.
+  //
+  llfs::PageCacheOvercommit overcommit;
+  overcommit.allow(true);
+  BATT_CHECK(!overcommit.is_triggered());
+
   // A MemTable has filled up.
   //
   if (delta_batch) {
     // Apply the finalized MemTable to the current checkpoint (in-memory).
     //
-    StatusOr<usize> push_status =
-        TURTLE_KV_COLLECT_LATENCY(this->metrics_.apply_batch_latency,
-                                  this->checkpoint_generator_.apply_batch(std::move(delta_batch)));
+    StatusOr<usize> push_status = TURTLE_KV_COLLECT_LATENCY(
+        this->metrics_.apply_batch_latency,
+        this->checkpoint_generator_.apply_batch(std::move(delta_batch), overcommit));
 
     BATT_REQUIRE_OK(push_status);
     BATT_CHECK_EQ(*push_status, 1);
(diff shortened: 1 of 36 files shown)


src/turtle_kv/tree/packed_leaf_page.hpp
```diff
@@ -735,18 +737,37 @@ inline PackedLeafPage* build_leaf_page(MutableBuffer buffer,
   if (plan.trie_index_reserved_size > 0) {
     const MutableBuffer trie_buffer{(void*)advance_pointer(buffer.data(), plan.trie_index_begin),
                                     plan.trie_index_end - plan.trie_index_begin};
+
     usize step_size = plan.compute_trie_step_size();
+    if (plan.key_count <= step_size) {
+      step_size = 1;
+    } else if (plan.key_count < 256) {
+      step_size = plan.key_count / 16;
+    }
```
vidyasilai: I think we could put this logic within `compute_trie_step_size`.

src/turtle_kv/kv_store.cpp
```diff
@@ -1056,10 +1068,19 @@ StatusOr<std::unique_ptr<CheckpointJob>> KVStore::apply_batch_to_checkpoint(
 
   // If the batch count is below the checkpoint distance, we are done.
   //
-  if (!this->should_create_checkpoint()) {
+  if (!overcommit.is_triggered() && !this->should_create_checkpoint()) {
     return nullptr;
   }
 
+  if (overcommit.is_triggered()) {
```
vidyasilai, line 1075: I'm wondering if adding a counter to the KVStore metrics to track how many times we hit this case where the overcommit is triggered would be useful here?

src/turtle_kv/tree/subtree_viability.hpp
```diff
@@ -44,6 +44,11 @@ inline std::ostream& operator<<(std::ostream& out, const NeedsMerge& t)
 //=#=#==#==#===============+=+=+=+=++=++++++++++++++-++-+--+-+----+---------------
 //
 struct NeedsSplit {
+  u16 pivot_count = 0;
+  u16 segment_count = 0;
+  i16 height = 0;
```
vidyasilai, line 49: Would you mind clarifying why these fields were added? Is it for debugging purposes for printing in the `operator<<` below?

vidyasilai: Just realized I commented a little too soon on this (oops!) without seeing that height is used in `normal_flush_might_fix`.

src/turtle_kv/tree/subtree.cpp
```diff
@@ -257,13 +261,13 @@ Status Subtree::apply_batch_update(const TreeOptions& tree_options,
...
 //
-Status Subtree::split_and_grow(BatchUpdateContext& context,
+Status Subtree::split_and_grow(BatchUpdateContext& update_context,
                                const TreeOptions& tree_options,
                                const KeyView& key_upper_bound)
 {
   BATT_CHECK(!this->locked_.load());
 
-  StatusOr<Optional<Subtree>> upper_half_subtree = this->try_split(context);
+  StatusOr<Optional<Subtree>> upper_half_subtree = this->try_split(update_context);
```
vidyasilai, line 269: With the new refactor of `try_split` in `InMemoryNode`, I believe the code on lines 234-238 of this file can be removed.

src/turtle_kv/tree/in_memory_node.cpp
```diff
@@ -979,6 +1008,83 @@ bool InMemoryNode::is_viable(IsRoot is_root) const
...
+      if (can_reflush) {
+        BATT_REQUIRE_OK(this->flush_to_pivot(context, reflush_pivot_i));
+        continue;
+      }
+    }
+
+    // If max flush factor is >1 and we are at the bottom of the tree, then we can attempt a normal
+    // flush, so long as the number of pivots is under the split limit.
+    //
+    if (this->tree_options.max_flush_factor() > 1 && normal_flush_count < this->pivot_count() &&
```
vidyasilai, line 1059: Is there a reason why we don't try flushing before trying to compact the levels if the height is 2? Is it more costly than compaction here?

Additionally, is the reason why we don't try a flush (at any node height) before calling `try_split_direct` because we want to avoid cascading flushes?

vidyasilai (commented):
Looks great! Just a few small comments/questions.

src/turtle_kv/tree/in_memory_node.cpp
tonyastolfi: Exactly right; compactions local to a node, while not always "cheap" in the sense that they might cause I/O, are bounded; we do want to avoid flushing if at all possible because of the possibility of cascades.

src/turtle_kv/tree/packed_leaf_page.hpp
tonyastolfi: Done.

src/turtle_kv/tree/subtree.cpp
tonyastolfi: Interesting...

After digging into this a bit I think I disagree.  The reason is that the new `InMemoryNode::try_split` doesn't exactly have a case like the "extra flush" inside `Subtree::apply_batch_update`, before the call to `split_and_grow`.  It does trigger more flushes to the most recently flushed pivot (`try_flush` will pick the one with most pending bytes), and it also flushes at the bottom of the tree when `normal_flush_might_fix(viability)` returns true.  This last case is almost but not quite the same as post-batch "extra flush," which is triggered not when `height == 2`, but at the root.

I will refactor this code a little to make the distinction more clear, but I think for now we should keep the existing logic.

src/turtle_kv/tree/subtree_viability.hpp
tonyastolfi: Yes, for diagnosis (human) and for identifying special cases so we can implement heuristic strategies for certain edge cases.

src/turtle_kv/kv_store.cpp
tonyastolfi: Good idea!  Added.

src/turtle_kv/tree/in_memory_node.cpp
vidyasilai: Thank you for the explanation!

src/turtle_kv/tree/subtree.cpp
vidyasilai: Thank you for the explanation, the small new refactor helps me see this now.

vidyasilai (approved):

PR #9, opened 2026-02-02
Introduce PiecewiseFilter for Update Buffer Segments

vidyasilai:
Introduces the `PiecewiseFilter` class into update buffer segments, replacing the flushed pivot bit set and flushed upper bound vector.


src/turtle_kv/util/piecewise_filter.hpp
```diff
@@ -0,0 +1,143 @@
...
+
+  /** \brief Constructs a default instance of a PiecewiseFilter object, initialized with no item
+   * range and filtered items.
+   */
+  PiecewiseFilter() noexcept;
+
+  /** \brief Filters out the item range provided by the specified interval of item indexes.
+   *
+   * \param i The item index range to filter out, specified as a half open interval.
+   * \param total_items The total number of items in the range.
```
tonyastolfi: (Note to self) When/why is `total_items` different from `i.size()`?

src/turtle_kv/util/piecewise_filter.hpp
```diff
@@ -0,0 +1,143 @@
...
+   */
+  void drop_index_range(Interval<OffsetT> i, OffsetT total_items);
+
+  /** \brief Returns whether or not the item at index `i` has been filtered out.
+   *
+   * \param i The item index being queried.
+   * \param total_items The total number of items in the range.
```
tonyastolfi: It's not clear to me which "range" this is referring to.

src/turtle_kv/util/piecewise_filter.hpp
```diff
@@ -0,0 +1,143 @@
...
+
+  /** \brief Returns the next unfiltered index closest to `i`.
+   *
+   * \param i The item index being queried.
+   * \param total_items The total number of items in the range.
+   *
+   * \return An index representing the next unfiltered item closest to index `i`. If `i` itself is
+   * unfiltered, `i` is returned. If everything past `i` is filtered, `None` is returned.
+   */
+  Optional<OffsetT> next_live_index(OffsetT i, OffsetT total_items) const;
```
tonyastolfi: Name suggests that this should advance `i` to the next live index *after* `i`; suggest renaming to something like `live_lower_bound`.

src/turtle_kv/util/piecewise_filter.hpp
```diff
@@ -0,0 +1,143 @@
...
+
+  /** \brief Finds an unfiltered slice of items starting at item index `start_i`.
+   *
+   * \param start_i The starting item for the range being queried.
+   * \param total_items The total number of items in the range.
+   *
+   * \return A half open interval representing the next slice of unfiltered item indexes starting
+   * from index `start_i`. If no such interval exists, or `start_i` is filtered, `None` is returned.
+   */
+  Optional<Interval<OffsetT>> next_live_interval(OffsetT start_i, OffsetT total_items) const;
```
tonyastolfi: Based on my comment for `next_live_index`, suggest naming this one something like `find_live_range`; we could have a matched pair of `live_lower_bound` and `live_upper_bound`, but I think this may not be a good way to get the effect of (what is currently) `next_live_interval`, since it may end up duplicating some work.

src/turtle_kv/tree/algo/segmented_levels.hpp
```diff
@@ -196,52 +192,22 @@ struct SegmentedLevelAlgorithms {
...
-      auto pivot_last = leaf_view.lower_bound(pivot_upper_bound_key);
-      auto flushed_last = leaf_view.lower_bound(max_key);
+      const PackedLeafPage& leaf = PackedLeafPage::view_of(pinned_page.get_page_buffer());
 
-      if (flushed_last != leaf_view.items_end() && get_key(*flushed_last) <= max_key) {
-        ++flushed_last;
-      }
-      BATT_CHECK((flushed_last == leaf_view.items_end()) || (get_key(*flushed_last) > max_key));
+      usize pivot_first =
+          std::distance(leaf.items_begin(), leaf.lower_bound(pivot_lower_bound_key));
```
tonyastolfi: nit: suggest finding some way to shorten this line so that the symmetry between it and the next is more evident, instead of there being a line break after the ` = `.

src/turtle_kv/tree/algo/segmented_levels.hpp
```diff
@@ -196,52 +192,22 @@ struct SegmentedLevelAlgorithms {
...
+      segment.set_filter_items(leaf.items_slice());
+      segment.drop_key_range(flush_key_crange);
 
-      // Check to see whether we are flushing all keys in the pivot range.
-      //
-      if (flushed_last == pivot_last) {
-        segment.set_flushed_item_upper_bound(pivot_i, 0);
+      Optional<u32> next_live_item =
+          segment.next_live_item(this->level_, leaf.key_count, BATT_CHECKED_CAST(u32, pivot_first));
+      if (!next_live_item || *next_live_item >= BATT_CHECKED_CAST(u32, pivot_last)) {
```
tonyastolfi: I'd like to suggest a cheaper way to achieve this; since we must pay to look up the first and last keys of the dropped range within the leaf, and we know the pivot bounding keys (`pivot_lower_bound_key`, `pivot_upper_bound_key`), why not just compare the key _after_ the leaf's lower bound of `flush_key_crange.upper_bound` against `pivot_upper_bound_key`?  If it is `>=`, then we drop the pivot.     This avoids two extra lookups in the leaf (for the pivot keys).  

Also, if we are only going to check the upper bound of the dropped range here, then we might as well not change the signature of this function from what it was originally; the implication is that `flush_key_crange.lower_bound` is always not-greater-than `pivot_lower_bound_key`.  If we want to keep the more general signature (passing the `CInterval` plus pivot instead of max key plus pivot), then we need to do an additional check here to make sure that the lower bound dropped range is not-after `pivot_first`.   If we go with the more general signature (I think I slightly prefer this option), then we can have `drop_key_range` return an interval of offsets, corresponding to the _new_ matched, dropped range.  This shouldn't cost any extra work inside `drop_key_range`, as we pretty much need to figure it out anyhow.

src/turtle_kv/tree/algo/segmented_levels.hpp
```diff
@@ -307,7 +273,10 @@ struct SegmentedLevelAlgorithms {
 
       BATT_CHECK_LE(pivot_offset_in_leaf, split_offset_in_leaf);
 
-      BATT_CHECK(in_segment(segment).split_pivot(pivot_i, split_offset_in_leaf, this->level_));
+      BATT_CHECK(in_segment(segment).split_pivot(pivot_i,
+                                                 split_offset_in_leaf,
+                                                 leaf_page.key_count,
```
tonyastolfi: I think if it works out to _not_ pass in the offset/index bound to various `PiecewiseFilter` functions (as we discussed in person), then we shouldn't need to pass `leaf_page.key_count` here, right?

src/turtle_kv/util/piecewise_filter.hpp
vidyasilai: Done

src/turtle_kv/util/piecewise_filter.hpp
vidyasilai: Done

src/turtle_kv/tree/algo/segmented_levels.hpp
vidyasilai: Yes, we shouldn't need to pass the total item count here. I fixed the function signature in `segments.hpp` to reflect this.

src/turtle_kv/util/piecewise_filter.hpp
```diff
@@ -43,57 +39,31 @@ class PiecewiseFilter
...
    *
    * \param i The item index being queried.
-   * \param total_items The total number of items in the range.
+   * \param total_items The total number of items.
    *
    * \return An index representing the next unfiltered item closest to index `i`. If `i` itself is
    * unfiltered, `i` is returned. If everything past `i` is filtered, `None` is returned.
    */
-  Optional<OffsetT> next_live_index(OffsetT i, OffsetT total_items) const;
+  Optional<OffsetT> live_lower_bound(OffsetT i, OffsetT total_items) const;
```
tonyastolfi: I think it's better to remove the `total_items` param, and return `OffsetT` instead.

src/turtle_kv/util/piecewise_filter.ipp
tonyastolfi: Suggest writing this to be more similar to `live_lower_bound` ... find the live lower bound; if that is greater than the input upper bound, you can just return an empty interval.  Otherwise, we can look at the _next_ dropped range (if `std::next(iter) != end`), and adjust the input end to be `min(end_i, next(iter)->lower_bound)`. 

We should return `Interval` instead of optional; return an empty interval in the `None` case.

tonyastolfi: Reminder: this should be happening outside the call (removing the need to pass `total_items`)

src/turtle_kv/util/piecewise_filter.ipp
```diff
@@ -0,0 +1,191 @@
...
+bool PiecewiseFilter<OffsetT>::live_at_index(OffsetT i) const
+{
+  auto iter = std::lower_bound(this->dropped_.begin(),
+                               this->dropped_.end(),
+                               i,
+                               typename Interval<OffsetT>::LinearOrder{});
+  if (iter != this->dropped_.end() && iter->contains(i)) {
+    return false;
+  }
+  return true;
```
tonyastolfi: Maybe simplify this function to just: `return this->live_lower_bound(i) == i;`

src/turtle_kv/util/piecewise_filter.test.cpp
```diff
@@ -0,0 +1,236 @@
...
+  filter = PiecewiseFilter<usize>::from_dropped(as_slice(dropped));
+  EXPECT_FALSE(filter.ok());
+  EXPECT_EQ(filter.status(), Status{::batt::StatusCode::kInvalidArgument});
+}
+
+//==#==========+==+=+=++=+++++++++++-+-+--+----- --- -- -  -  -   -
+//
+TEST(PiecewiseFilterTest, QueryTest)
+{
+  std::default_random_engine rng{/*seed=*/3};
```
tonyastolfi: We should loop over many seeds here to get better test coverage.

src/turtle_kv/util/piecewise_filter.test.cpp
```diff
@@ -0,0 +1,236 @@
...
+//
+TEST(PiecewiseFilterTest, KeyQueryTest)
+{
+  u32 num_keys = 1000;
+  std::vector<std::string_view> keys;
+  keys.reserve(num_keys);
+
+  // Generate some random strings and sort them.
+  //
+  std::default_random_engine rng{30};
```
tonyastolfi: Also, good here to iterate for more seeds.

src/turtle_kv/util/piecewise_filter.ipp
```diff
@@ -0,0 +1,191 @@
...
+
+//==#==========+==+=+=++=+++++++++++-+-+--+----- --- -- -  -  -   -
+//
+template <typename OffsetT>
+/*static*/ StatusOr<PiecewiseFilter<OffsetT>> PiecewiseFilter<OffsetT>::from_dropped(
+    const Slice<const Interval<OffsetT>>& dropped)
+{
+  PiecewiseFilter<OffsetT> filter;
+
+  filter.dropped_total_ = 0;
```
tonyastolfi: Suggest breaking this block of code out into a `check_invariants` method that  can be used during tests.

src/turtle_kv/tree/algo/segments.hpp
```diff
@@ -151,25 +141,46 @@ struct SegmentAlgorithms {
...
         });
+
+    return OkStatus();
   }
 
   /** \brief Searches the segment for the given key, returning its value if found.
    */
   template <typename LevelT>
-  batt::seq::LoopControl find_key(LevelT& level,
-                                  i32 key_pivot_i,
```
tonyastolfi, line 170: Let's leave `key_pivot_i` in the signature for now

src/turtle_kv/util/piecewise_filter.ipp
vidyasilai: Done

src/turtle_kv/util/piecewise_filter.test.cpp
vidyasilai: Done

src/turtle_kv/util/piecewise_filter.test.cpp
vidyasilai: Done

src/turtle_kv/util/piecewise_filter.ipp
vidyasilai: Done

src/turtle_kv/tree/algo/segments.hpp
vidyasilai: Done

PR #10, opened 2026-02-05
Recover ChangeLogBlocks from ChangeLogFile

gabrielbornstein:
Includes new functions to recover and initialize ChangeLogBlocks from the ChangeLogFile after the KVStore has been shutdown, and the ChangeLogFile has been written to disk.

Also includes refactoring for the KVStore test suite, so that there is less redundant code, and future tests will be more portable and easier to write.

Note: Some investigation is needed. ChangeLogBlock::verify() is failing, however, we are reading correct and consistent ChangeLogBlock data.


conanfile.py
```diff
@@ -56,23 +54,19 @@ def requirements(self):
...
         self.requires("abseil/20250127.0", **VISIBLE, **OVERRIDE)
-        self.requires("batteries/[>=0.61.0 <1]", **VISIBLE, **OVERRIDE)
+        self.requires("batteries/0.60.2", **VISIBLE, **OVERRIDE)
         self.requires("boost/1.88.0", **VISIBLE, **OVERRIDE)
         self.requires("glog/0.7.1", **VISIBLE)
-        self.requires("llfs/[>=0.43.3 <1]", **VISIBLE)
+        self.requires("gperftools/2.16", **VISIBLE)
+        self.requires("llfs/0.42.0", **VISIBLE)
         self.requires("pcg-cpp/cci.20220409", **VISIBLE)
+        self.requires("vqf/0.2.5-devel", **VISIBLE)
```
gabrielbornstein: Remove -devel before merging

src/turtle_kv/change_log_block.hpp
```diff
@@ -77,9 +82,28 @@ class ChangeLogBlock
...
+  i32 ref_count() noexcept
   {
-    return this->ref_count_;
+    return this->ephemeral_state_ptr()->ref_count_;
+  }
+
+  /** \brief Helper function to initialize the ephemeral state of this ChangeLogBlock. Transfers
+   * ownership of grant to ChangeLogBlock, and initializes the reference count to ref_count.
+   */
+  void init_ephemeral_state(batt::Grant& grant, i64 ref_count = 1)
```
gabrielbornstein: Pass grant as rvalue ref (&&)

src/turtle_kv/change_log_file.cpp
```diff
@@ -146,6 +146,15 @@ auto ChangeLogFile::PackedConfig::unpack() const noexcept -> ChangeLogFile::Conf
 //
 ChangeLogFile::~ChangeLogFile() noexcept
 {
+  Interval<i64> block_range{0, this->size()};
```
gabrielbornstein: use this->active_blocks() instead

src/turtle_kv/change_log_file.cpp
```diff
@@ -231,6 +240,64 @@ StatusOr<batt::Grant> ChangeLogFile::reserve_blocks(
   return grant;
 }
 
+//==#==========+==+=+=++=+++++++++++-+-+--+----- --- -- -  -  -   -
+//
+ChangeLogFile::ReadLock ChangeLogFile::acquire_read_lock(batt::Grant& grant,
```
gabrielbornstein: acquire_read_lock - > set_block_range_in_use

src/turtle_kv/change_log_file.cpp
```diff
@@ -231,6 +240,64 @@ StatusOr<batt::Grant> ChangeLogFile::reserve_blocks(
...
+                                                         const Interval<i64>& block_range,
+                                                         i64 blocks_written) noexcept
+{
+  this->lock_for_read(block_range);
+
+  // Important: only do this after locking the range.
+  //
+  {
+    this->upper_bound_.fetch_add(blocks_written);
```
gabrielbornstein: refactor blocks_written -> block_range.size()

src/turtle_kv/change_log_file.cpp
gabrielbornstein: What if we just assert number of active_blocks is 0 instead? Isn't the expectation at this point that there are no active blocks, so we should never have to iterate over the blocks?

src/turtle_kv/change_log_block.hpp
```diff
@@ -183,6 +207,12 @@ class ChangeLogBlock
   /** \brief The members of this object which live outside the block buffer.
    */
   struct EphemeralState {
+    /** \brief Atomic reference counter to manage the lifetime of the buffer.
+     */
+    std::atomic<i32> ref_count_;
```
gabrielbornstein: Move ref_count back into header

src/turtle_kv/change_log_file.hpp
```diff
@@ -210,6 +234,60 @@ class ChangeLogFile
...
+    batt::MutableBuffer block_buffer{block_memory, static_cast<size_t>(this->config_.block_size)};
+
+    status = this->file_.read_all(curr_file_offset, block_buffer);
+
+    batt::StatusOr<batt::Grant> buffer_grant =
+        this->reserve_blocks(BlockCount{1}, batt::WaitForResource::kFalse);
+
+    BATT_REQUIRE_OK(buffer_grant);
+
+    block->init_ephemeral_state(std::move(*buffer_grant));
```
gabrielbornstein: Delay initializing EphemeralState

src/turtle_kv/change_log_block.hpp
```diff
@@ -270,21 +300,22 @@ class ChangeLogBlock
...
-  std::atomic<i32> ref_count_;  // TODO [tastolfi 2025-12-16] move to ephemeral state
-
-  // Pad the next field (this->next_) out to (void*) this + 32 bytes;
+  // TODO: [Gabe Bornstein 2/3/26] Whoever constructs a Mutable ChangeLogBlock should pass this
+  // value. Whoever creates a new MemTable also needs to pass this value to the MemTable
+  // construction. Needs to be added to MemTable. Open questions if this value should live in
+  // MemTable or a level below.
   //
-  u8 padding1_[4];
+  u32 prev_epoch_num_slots_;
```
gabrielbornstein: Delete this

src/turtle_kv/change_log_file.cpp
gabrielbornstein: Done

src/turtle_kv/change_log_block.hpp
gabrielbornstein: Done

src/turtle_kv/change_log_block.hpp
gabrielbornstein: Done

src/turtle_kv/change_log_file.hpp
gabrielbornstein: Not done. ephemeral state needs to be initialized so we don't get a nullptr dereference during destruction.

src/turtle_kv/change_log_block.hpp
gabrielbornstein: Done

src/turtle_kv/change_log_file.cpp
gabrielbornstein: Done

src/turtle_kv/change_log_file.hpp
```diff
@@ -119,6 +120,23 @@ class ChangeLogFile
...
+   * the returned blocks (https://www.boost.org/doc/libs/1_40_0/libs/smart_ptr/intrusive_ptr.html).
+   */
+  batt::StatusOr<std::vector<boost::intrusive_ptr<ChangeLogBlock>>> read_blocks_into_vector();
+
+  // TODO: [Gabe Bornstein 1/20/26] Consider using concepts here to define required parameters and
+  // return types?
+  //
+  /** \brief Read over all the blocks currently in the ChangeLogFile, calling process_block for each
+   * block. process_block is responsible for determining when to stop reading blocks, and what to do
+   * with each recovered block.
```
tonyastolfi: We should explicitly say that ownership of the memory of `ChangeLogBlock`s is transferred to `process_block`.

src/turtle_kv/change_log_file.hpp
```diff
@@ -119,6 +120,23 @@ class ChangeLogFile
...
+   */
+  template <typename SerializeFn = void(ChangeLogBlock*)>
```
tonyastolfi: Please update signature (default type) to `Status (ChangeLogBlock*)`

src/turtle_kv/change_log_file.hpp
```diff
@@ -210,6 +234,65 @@ class ChangeLogFile
...
+
+    ChangeLogBlock* block = reinterpret_cast<ChangeLogBlock*>(block_memory);
+
+    // TODO: [Gabe Bornstein 2/4/26] Investigate why verify is failing.
+    //
+    // block->verify();
+
+    // Create MutableBuffer for reading from file
+    //
+    batt::MutableBuffer block_buffer{block_memory, static_cast<size_t>(this->config_.block_size)};
```
tonyastolfi: nit: Please use `usize` in place of `size_t`

src/turtle_kv/change_log_file.hpp
```diff
@@ -210,6 +234,65 @@ class ChangeLogFile
...
+    batt::StatusOr<batt::Grant> buffer_grant =
+        this->reserve_blocks(BlockCount{1}, batt::WaitForResource::kFalse);
+
+    BATT_REQUIRE_OK(buffer_grant);
+
+    block->init_ephemeral_state(std::move(*buffer_grant));
+
+    Interval<i64> block_range{blocks_read, blocks_read + 1};
+
+    block->set_read_lock(set_block_range_in_use(block->get_grant(), block_range));
```
tonyastolfi: `this->set_block_range_in_use`

src/turtle_kv/change_log_block.cpp
```diff
@@ -30,21 +41,20 @@ namespace turtle_kv {
...
                                             batt::Grant&& grant,
                                             usize block_size) noexcept
-    : magic_{reinterpret_cast<u64>(this) ^ ChangeLogBlock::kMagic}
+    : magic_{ChangeLogBlock::kMagic}
     , owner_id_{owner_id}
     , block_size_{BATT_CHECKED_CAST(u16, block_size)}
     , slot_count_{0}
     , space_{BATT_CHECKED_CAST(u16,
                                this->block_size_ - (sizeof(ChangeLogBlock) + sizeof(SlotInfo)))}
-    , ref_count_{1}
```
tonyastolfi, line 39: Reminder: restore this initialization

src/turtle_kv/change_log_file.hpp
```diff
@@ -210,6 +234,65 @@ class ChangeLogFile
...
+
+    block->set_ref_count(1);
```
tonyastolfi: Once the initialization of `ref_count_` inside `ChangeLogBlock`'s ctor is restored, we should be able to remove this line.

src/turtle_kv/change_log_file.hpp
```diff
@@ -210,6 +234,65 @@ class ChangeLogFile
...
+
+    // process_block is responsible for determining when to stop reading. Usually, break once we've
+    // read a block that isn't valid or if we've reached the end of the file.
+    //
+    if (!process_block(block).ok()) {
```
tonyastolfi: I think we should save the returned status, and check it explicitly against `batt::StatusCode::kLoopBreak`; return the (failing) error code if not ok and not loop break.

src/turtle_kv/change_log_file.hpp
```diff
@@ -210,6 +234,65 @@ class ChangeLogFile
...
+    //
+    i64 curr_file_offset = this->config_.block0_offset + curr_block_offset;
+
+    void* block_memory = ChangeLogBlock::allocate_aligned(this->config_.block_size);
+
+    ChangeLogBlock* block = reinterpret_cast<ChangeLogBlock*>(block_memory);
+
+    // TODO: [Gabe Bornstein 2/4/26] Investigate why verify is failing.
+    //
+    // block->verify();
```
tonyastolfi: We should do this after reading the block data.

src/turtle_kv/change_log_file.cpp
```diff
@@ -231,6 +240,69 @@ StatusOr<batt::Grant> ChangeLogFile::reserve_blocks(
...
+    batt::Grant& grant,
+    const Interval<i64>& block_range) noexcept
+{
+  i64 blocks_written = block_range.size();
+  this->lock_for_read(block_range);
+
+  // Important: only do this after locking the range.
+  //
+  {
+    this->upper_bound_.fetch_add(blocks_written);
```
tonyastolfi: Let's not update `upper_bound_` here; the latest recovery design doc describes how we recover the active range.  I've confirmed by searching through the code that it would be safe to update `upper_bound_` outside this function, after it returns (at least in `ChangeLogFile::append`).

src/turtle_kv/change_log_file.hpp
gabrielbornstein: When we read a ChangeLogBlock from the ChangeLogFile, ref_count's are initialized to two. If we want them to be initialized to one, this set is necessary.

src/turtle_kv/change_log_file.hpp
gabrielbornstein: Done

src/turtle_kv/change_log_file.hpp
gabrielbornstein: Done

src/turtle_kv/change_log_file.hpp
gabrielbornstein: Done

src/turtle_kv/change_log_file.hpp
gabrielbornstein: Done

src/turtle_kv/change_log_block.cpp
gabrielbornstein: Done

src/turtle_kv/change_log_file.hpp
gabrielbornstein: Done

src/turtle_kv/change_log_file.cpp
gabrielbornstein: I tried this, it appears we can have active_blocks during destruction, however, their counter is 0. We need to iterate over blocks that appear active and verify count is 0. Is this a bug though? Don't we expect there to be 0 active blocks during destruction of the ChangeLogFile?

src/turtle_kv/change_log_block.hpp
```diff
@@ -53,6 +53,11 @@ class ChangeLogBlock
    */
   void operator delete(void* ptr) noexcept = delete;
 
+  /** \brief Allocates and returns a pointer of the specifed size aligned to the ChangeLogBlock
+   * default by alignment.
```
tonyastolfi: Awkward; suggest: `... aligned to ChangeLogBlock::kDefaultAlign bytes.`

src/turtle_kv/change_log_block.hpp
```diff
@@ -77,11 +84,33 @@ class ChangeLogBlock
    */
   void remove_ref(i32 count) noexcept;
 
-  i32 ref_count() const noexcept
+  i32 ref_count() noexcept
```
tonyastolfi: Why remove the `const` here?

src/turtle_kv/change_log_block.hpp
```diff
@@ -77,11 +84,33 @@ class ChangeLogBlock
...
   {
     return this->ref_count_;
   }
 
+  void set_ref_count(i64 ref_count = 1)
```
tonyastolfi: I'd recommend against having a default value for this arg.

src/turtle_kv/change_log_block.hpp
```diff
@@ -165,7 +194,7 @@ class ChangeLogBlock
 
   /** \brief Perform basic sanity checks to make sure this is a valid ChangeLogBlock object.
    */
-  void verify() const noexcept;
+  batt::Status verify() const noexcept;
```
tonyastolfi, line 262: We should also have a way to do a more thorough verification by recomputing the xxh3 hash and comparing against the value stored in the header.  Maybe another member function, `verify_data()`?

src/turtle_kv/change_log_block.hpp
```diff
@@ -248,15 +279,14 @@ class ChangeLogBlock
   }
 
   //+++++++++++-+-+--+----- --- -- -  -  -   -
-
```
tonyastolfi, line 251: Please undo.

src/turtle_kv/change_log_file.cpp
```diff
@@ -231,6 +237,64 @@ StatusOr<batt::Grant> ChangeLogFile::reserve_blocks(
...
+batt::StatusOr<std::vector<boost::intrusive_ptr<ChangeLogBlock>>>
+ChangeLogFile::read_blocks_into_vector()
+{
+  std::vector<boost::intrusive_ptr<ChangeLogBlock>> blocks;
+  batt::Status read_blocks_status = this->read_blocks([&](ChangeLogBlock* block) -> batt::Status {
+    BATT_CHECK_EQ(block->ref_count(), 1);
+
+    // If block size is zero, block has not been initialized. Stop reading here.
+    //
+    if (block->block_size() == 0) {
```
tonyastolfi: Maybe at this layer we can just `BATT_CHECK` that this is not the case?  I think the lower-level read blocks function should be responsible for determining that the true extents of recoverable blocks are, and which blocks are recoverable within that range.

src/turtle_kv/change_log_file.cpp
```diff
@@ -231,6 +237,64 @@ StatusOr<batt::Grant> ChangeLogFile::reserve_blocks(
...
+      // Need to free memory of untracked block here otherwise we will have a memory leak.
+      //
+      block->remove_ref(1);
+      return batt::StatusCode::kOutOfRange;
+    }
+    // TODO: [Gabe Bornstein 1/28/26] Decide if we want to keep block, decrement ref count if not.
+    //
+
+    blocks.push_back(boost::intrusive_ptr<ChangeLogBlock>{block, false});
```
tonyastolfi: Slightly better to use `emplace_back` here, to elide a temporary copy of the `intrusive_ptr`.

src/turtle_kv/change_log_file.hpp
gabrielbornstein: Done

src/turtle_kv/change_log_file.cpp
gabrielbornstein: Done

src/turtle_kv/change_log_file.cpp
gabrielbornstein: The idea of this check is to return early if we reach uninitialized blocks in the changeLogFile. Is there something better we could check for that? I updated the code so that we return kLoopBreak instead here.

src/turtle_kv/change_log_file.cpp
gabrielbornstein: Done

src/turtle_kv/change_log_block.hpp
gabrielbornstein: Done

src/turtle_kv/change_log_block.hpp
gabrielbornstein: Not sure... maybe it was relevant to something I undid. It's back

src/turtle_kv/change_log_block.hpp
gabrielbornstein: Done

src/turtle_kv/change_log_block.hpp
gabrielbornstein: Haven't done this yet. Do we want to add this during this MR, or can we delay it? Also, where would we want to call verify_data()? If it's necessary for checks for the recover architecture plan, lets add a TODO and add it during a subsequent MR.

src/turtle_kv/change_log_block.hpp
```diff
@@ -53,6 +53,11 @@ class ChangeLogBlock
...
   void operator delete(void* ptr) noexcept = delete;
 
+  /** \brief Allocates and returns a pointer of the specifed size aligned to
+   * ChangeLogBlock::kDefaultAlign bytes.
+   */
+  static void* allocate_aligned(usize n_bytes) noexcept;
+
   /** \brief Allocates and returns a buffer of the specifed size.
    */
   static ChangeLogBlock* allocate(u64 owner_id, batt::Grant&& grant, usize n_bytes) noexcept;
```
tonyastolfi, line 128: Let's add a second factory method to ChangeLogBlock: `StatusOr<ChangeLogBlock*> recover(MutableBuffer buf, batt::Grant&& grant)`; that way we can remove or at least make private `set_ref_count` and `init_ephemeral_state`, and this will simplify the code inside `ChangeLogFile::read_blocks`.  `recover` should do data integrity checks and return an error status code if they don't pass.

src/turtle_kv/change_log_block.hpp
```diff
@@ -53,6 +53,11 @@ class ChangeLogBlock
    */
   void operator delete(void* ptr) noexcept = delete;
 
+  /** \brief Allocates and returns a pointer of the specifed size aligned to
+   * ChangeLogBlock::kDefaultAlign bytes.
+   */
+  static void* allocate_aligned(usize n_bytes) noexcept;
```
tonyastolfi: This should return `MutableBuffer`

src/turtle_kv/change_log_file.cpp
tonyastolfi: It would be good to return a status code from (to be added) `ChangeLogBlock::recover` that indicates this looks like a block which was never written; that way the code in `ChangeLogFile` can just stop there.

src/turtle_kv/change_log_block.hpp
gabrielbornstein: Make sure to return a batt::Status that denotes when we read an EOF block, or malformed block.

gabrielbornstein: Make sure that recover frees memory in case a block is malformed or EOF. Like a "free_aligned" function.

src/turtle_kv/change_log_block.hpp
tonyastolfi: I suppose we could defer, but `ChangeLogBlock::recover` would be a natural place for this check to go, and I don't think the MR is too unwieldy (yet anyhow ;-))

src/turtle_kv/change_log_block.hpp
```diff
@@ -53,17 +55,41 @@ class ChangeLogBlock
    */
   void operator delete(void* ptr) noexcept = delete;
```
tonyastolfi, line 120: We should add an RAII type to contain the allocated (aligned) buffer; something like:

```c++
class ChangeLogBlockMemory
{
public:
  using Self = ChangeLogBlockMemory;

  explicit ChangeLogBlockMemory(void* ptr, usize size) noexcept 
    : buffer_{ptr, size}
  {}

  ChangeLogBlockMemory(const Self& ) = delete;
  Self& operator=(const Self& ) = delete;

  ChangeLogBlockMemory(Self&& other) noexcept 
  {// TODO!
  }

  Self& operator=(Self&& other) noexcept 
  {// TODO!
  }

  ~ChangeLogBlockMemory() noexcept 
  {
    if (this->buffer_.data() != nullptr) {
       free(this->buffer_.data());
    }
  }

  void * data() const {
    return this->buffer_.data();
  }

  usize size() const {
    return this->buffer_.size();
  }

  void* release_ownership() {
     void* released_ptr = this->buffer_.data();
     this->buffer_ = MutableBuffer{};
     return released_ptr;
  }

private:
  MutableBuffer buffer_;
};
```

Then, 

```c++
ChangeLogBlockMemory ChangeLogBlock::allocate_aligned(usize size) noexcept
{
   return ChangeLogBlockMemory{std::allocate_aligned(size), size};
}

StatusOr<boost::intrusive_ptr<ChangeLogBlock>> ChangeLogBlock::recover(ChangeLogBlockMemory&& memory, Grant&& grant) 
{
   ChangeLogBlock* block = static_cast<ChangeLogBlock*>(memory.data());

   // do all our validations
   if (block->block_size() != memory.size()) {
      return {StatusCode::kWhatever};
   }

   // it looks... ok!
   return boost::intrusive_ptr<ChangeLogBlock>{ memory.release_ownership() };
} 
```

src/turtle_kv/change_log_block.cpp
```diff
@@ -149,10 +201,24 @@ batt::Grant ChangeLogBlock::consume_grant() noexcept
 
 //==#==========+==+=+=++=+++++++++++-+-+--+----- --- -- -  -  -   -
 //
-void ChangeLogBlock::verify() const noexcept
+batt::Status ChangeLogBlock::verify() const noexcept
+{
+  BATT_REQUIRE_NE(this->magic_, ChangeLogBlock::kExpired);
```
tonyastolfi: We could probably just replace this line with a `static_assert` that `kExpired` and `kMagic` aren't equal.

src/turtle_kv/change_log_block.hpp
```diff
@@ -48,22 +50,101 @@ class ChangeLogBlock
     u16 offset;
   };
 
+  class ChangeLogBlockMemory
```
tonyastolfi: Since this is a member type within `ChangeLogBlock` it seems a little redundant to repeat `CheckLogBlock` in the name... perhaps something like `Memory` or `ScopedMemory`?

src/turtle_kv/change_log_block.hpp
```diff
@@ -82,6 +163,28 @@ class ChangeLogBlock
     return this->ref_count_;
   }
 
+  void set_ref_count(i64 ref_count)
```
tonyastolfi: I think `set_ref_count` and `init_ephemeral_state` can become private, now that we have `recover`.

src/turtle_kv/change_log_block.hpp
tonyastolfi: Seems `verify_data` is now implemented?  Resolving

src/turtle_kv/change_log_file.hpp
```diff
@@ -210,6 +237,73 @@ class ChangeLogFile
...
+
+    batt::StatusOr<batt::Grant> buffer_grant =
+        this->reserve_blocks(BlockCount{1}, batt::WaitForResource::kFalse);
+
+    BATT_REQUIRE_OK(buffer_grant);
+
+    ChangeLogBlock::ChangeLogBlockMemory block_memory =
+        ChangeLogBlock::allocate_aligned(this->config_.block_size);
+
+    MutableBuffer block_buffer{block_memory.data(), block_memory.size()};
```
tonyastolfi: This works; consider perhaps adding a `buffer()` method to the memory type to make it a little cleaner?

gabrielbornstein: Gladly! Thought there might be a reason we were only exposing data() and size(), so didn't want to just expose buffer without your go ahead 👍

src/turtle_kv/change_log_block.cpp
gabrielbornstein: That doesn't make sense to me... Isn't this line checking that we haven't marked the contents of the ChangeLogBlock as Expired? It won't catch anything that `BATT_REQUIRE_EQ(this->magic_, ChangeLogBlock::kMagic);` wouldn't catch, but it'd give us more debugging information.

The destructor of ChangeLogBlock sets this->magic_ to kExpired on line ~111 of this file.

src/turtle_kv/change_log_block.hpp
gabrielbornstein: Done

src/turtle_kv/change_log_block.hpp
gabrielbornstein: Done

src/turtle_kv/change_log_block.cpp
tonyastolfi: Yes, you are right that we are checking that we haven't marked the object as expired (by setting `magic_` to `kExpired`), but it looks a bit funny to me that we are checking that the same field _is_ one thing and _is not_ another, one line after another.  It seems sufficient to me check what it _is_; what `magic_` is _not_ is then true by virtue of `kExpired != kMagic`, which can be checked at compile time, since both are `constexpr`.

Hope this makes it clearer.

src/turtle_kv/change_log_block.hpp
```diff
@@ -48,22 +50,106 @@ class ChangeLogBlock
...
+    {
+      return this->buffer_.data();
+    }
+
+    usize size() const
+    {
+      return this->buffer_.size();
+    }
+
+    MutableBuffer buffer()
```
tonyastolfi: Can be `const`.

src/turtle_kv/change_log_file.cpp
```diff
@@ -231,6 +237,60 @@ StatusOr<batt::Grant> ChangeLogFile::reserve_blocks(
...
+ChangeLogFile::read_blocks_into_vector()
+{
+  std::vector<boost::intrusive_ptr<ChangeLogBlock>> blocks;
+  batt::Status read_blocks_status =
+      this->read_blocks([&](boost::intrusive_ptr<ChangeLogBlock> block) -> batt::Status {
+        BATT_CHECK_EQ(block->ref_count(), 1);
+
+        // If block size is zero, block has not been initialized. Stop reading here.
+        //
+        if (block->block_size() == 0) {
```
tonyastolfi: I don't think this should ever be true; since `read_blocks` now calls `ChangeLogBlock::recover`, it has verified that all blocks which are passed on to this point have size equal to the size of the data read, which must be non-zero.

src/turtle_kv/change_log_file.cpp
```diff
@@ -231,6 +237,60 @@ StatusOr<batt::Grant> ChangeLogFile::reserve_blocks(
...
+          return batt::StatusCode::kLoopBreak;
+        }
+
+        blocks.push_back(block);
```
tonyastolfi, line 276: `blocks.emplace_back(std::move(block));`

src/turtle_kv/change_log_file.hpp
```diff
@@ -119,6 +120,26 @@ class ChangeLogFile
...
+  // return types?
+  //
+  /** \brief Read over all the blocks currently in the ChangeLogFile, calling process_block for each
+   * block.
+   * process_block is responsible for determining when to stop reading blocks, and what to do
+   * with each recovered block.
+   * Ownership of the ChangeLogBlock's memory is transferred to `process_block`.
+   * `process_block` must free the block's memory if it plans to do nothing with it.
+   */
+  template <typename SerializeFn = batt::Status(ChangeLogBlock*)>
```
tonyastolfi: Please update the signature to use `boost::intrusive_ptr`

src/turtle_kv/change_log_file.hpp
```diff
@@ -210,6 +237,71 @@ class ChangeLogFile
...
+
+    Interval<i64> block_range{blocks_read, blocks_read + 1};
+
+    (*block)->set_read_lock(this->set_block_range_in_use((*block)->get_grant(), block_range));
+
+    this->upper_bound_.fetch_add(block_range.size());
+
+    // `process_block` is responsible for determining when to stop reading.
+    //
+    batt::Status process_status = process_block(*block);
```
tonyastolfi: `... process_block(std::move(*block));`

src/turtle_kv/change_log_file.hpp
```diff
@@ -210,6 +237,71 @@ class ChangeLogFile
...
+    if (block.status() == batt::StatusCode::kOutOfRange ||
+        block.status() == batt::StatusCode::kDataLoss) {
+      LOG(INFO) << "Recovered " << blocks_read
+                << " blocks. Stopped reading with status:" << BATT_INSPECT(block.status());
+      return batt::OkStatus();
+    }
+
+    Interval<i64> block_range{blocks_read, blocks_read + 1};
+
+    (*block)->set_read_lock(this->set_block_range_in_use((*block)->get_grant(), block_range));
```
tonyastolfi: It is too soon in the recovery process to know what the logical block range is; I suggest removing this line (and 286), maybe add a TODO comment.

src/turtle_kv/change_log_file.hpp
```diff
@@ -210,6 +237,71 @@ class ChangeLogFile
...
+
+    this->upper_bound_.fetch_add(block_range.size());
```
tonyastolfi: Same comment here; too soon to know.  Suggest removing.

src/turtle_kv/kv_store.test.cpp
```diff
@@ -498,13 +590,13 @@ INSTANTIATE_TEST_SUITE_P(
...
         // TODO: [Gabe Bornstein 11/5/25] Investigate: We ARE
         // getting checkpoint data for this case. Does taking additional checkpoints flush keys?
-        CheckpointTestParams(/* num_puts */ 2, /* num_checkpoints_to_create */ 100),
-        CheckpointTestParams(/* num_puts */ 100, /* num_checkpoints_to_create */ 100),
-        CheckpointTestParams(/* num_puts */ 1, /* num_checkpoints_to_create */ 100000),
-        CheckpointTestParams(/* num_puts */ 1, /* num_checkpoints_to_create */ 0),
-        CheckpointTestParams(/* num_puts */ 0, /* num_checkpoints_to_create */ 100),
-        CheckpointTestParams(/* num_puts */ 5, /* num_checkpoints_to_create */ 100000),
-        CheckpointTestParams(/* num_puts */ 10, /* num_checkpoints_to_create */ 100000)
+        CheckpointTestParams(/* num_checkpoints_to_create */ 2, /* num_puts */ 100),
```
tonyastolfi: Consider removing the ctor of `CheckpointTestParams` and defining it as a simple `struct`, so you can use field initialization syntax, e.g. `CheckpointTestParams{.num_checkpoints_to_create=100, .num_puts = 100}`

src/turtle_kv/change_log_file.cpp
tonyastolfi: "Is this a bug...?"  Perhaps... would you do some more investigation to find out why the blocks with count 0 aren't removed from active?

src/turtle_kv/change_log_file.cpp
gabrielbornstein: I've noticed you tend to prefer emplace_back over push_back. I'm wondering why? I thought emplace_back was only faster if we had yet to construct the object, but here, we already have an instrusive_ptr to ChangeLogBlock. I was just reading this little discussion on push_back vs. emplace_back:
https://groups.google.com/a/chromium.org/g/chromium-dev/c/7mJypsYz6AA?pli=1

src/turtle_kv/change_log_block.cpp
gabrielbornstein: Gotcha, I'll do that. Maybe this is leftovers from some old debugging then? This is the current implementation for verify on main:
```
void ChangeLogBlock::verify() const noexcept
{
  BATT_CHECK_NE(this->magic_, ChangeLogBlock::kExpired);
  BATT_CHECK_EQ(this->magic_, reinterpret_cast<u64>(this) ^ ChangeLogBlock::kMagic);
}
```

src/turtle_kv/change_log_file.cpp
gabrielbornstein: After some digging, it looks like we're skipping a lot of the calls that would update the active range. The only place `ChangeLogFile::lower_bound_` is updated is inside `ChangeLogFile::update_lower_bound`. If `this->read_lock_counter_per_block_[addr]->load() > 0`, then we will skip updating the lower_bound to the new passed bound. We're seeing this happen frequently during ChangeLogFile destruction, including the case where the lower_bound would be updated to the largest known upper_bound.

Here's most of the actual `update_lower_bound` function:
```i64 observed_lower_bound = this->lower_bound_.load();
    i64 new_lower_bound = observed_lower_bound;
    old_lower_bound = observed_lower_bound;

    i64 addr = new_lower_bound % this->config_.block_count;
    while (new_lower_bound < update_upper_bound) {
      if (this->read_lock_counter_per_block_[addr]->load() > 0) {
        break;
      }
      ++addr;
      if (addr == this->config_.block_count) {
        addr = 0;
      }
      ++new_lower_bound;
      ++n_blocks_freed;
    }

    LOG(INFO) << "ChangeLogFile::update_lower_bound observed_lower_bound: " << observed_lower_bound
              << ", new_lower_bound: " << new_lower_bound;

    while (observed_lower_bound < new_lower_bound) {
      if (this->lower_bound_.compare_exchange_weak(observed_lower_bound, new_lower_bound)) {
        break;
      }
    }
```

This function gets called via the path ~ChangeLogReadLock() -> ChangeLogFile::unlock_for_read -> ChangeLogFile::update_lower_bound.

The value of `this->read_lock_counter_per_block_[addr]->load()` is 1 when we skip updating the lower_bound.
It looks like we decrement  `this->read_lock_counter_per_block_[addr]` in `ChangeLogFile::unlock_for_read` before calling `update_lower_bound`, during ChangeLogFile destruction, but it isn't getting decremented to zero, hence we aren't updating the active range.

gabrielbornstein: It looks like we don't destruct  ChangeLogReadLocks in order. The order is pretty random. Because of this, we decrement the read_lock_counter out of order. When we calculate `addr` in `update_lower_bound`, we calculate the same block address until we increment `ChangeLogFile::lower_bound_`. This means our check, ` if (this->read_lock_counter_per_block_[addr]->load() > 0)` gets stuck on one block until we decrement that blocks counter to zero. Only then can we move on to the next block and update the lower bound.

tldr; Our implementation depends on ChangeLogReadLocks getting destructed in incrementing order but they aren't.

here's some example output of the order ReadLocks are getting destructed in:
```
I20260311 14:15:48.471795 140687971209664 change_log_read_lock.cpp:41] ~ChangeLogReadLock() on range  this->block_range_ == [1336,1337)
I20260311 14:15:48.471832 140687971209664 change_log_read_lock.cpp:41] ~ChangeLogReadLock() on range  this->block_range_ == [1335,1336)
I20260311 14:15:48.471895 140687971209664 change_log_read_lock.cpp:41] ~ChangeLogReadLock() on range  this->block_range_ == [1334,1335)
I20260311 14:15:48.471895 140687971209664 change_log_read_lock.cpp:41] ~ChangeLogReadLock() on range  this->block_range_ == [1487,1488)
I20260311 14:15:48.471949 140687971209664 change_log_read_lock.cpp:41] ~ChangeLogReadLock() on range  this->block_range_ == [1486,1487)
```

src/turtle_kv/kv_store.test.cpp
gabrielbornstein: Done

src/turtle_kv/change_log_file.hpp
gabrielbornstein: Done

src/turtle_kv/change_log_file.hpp
gabrielbornstein: Done

src/turtle_kv/change_log_file.hpp
gabrielbornstein: Done

src/turtle_kv/change_log_file.hpp
gabrielbornstein: Done

src/turtle_kv/change_log_file.cpp
gabrielbornstein: Done

src/turtle_kv/change_log_block.hpp
gabrielbornstein: Done

src/turtle_kv/change_log_file.cpp
gabrielbornstein: We fixed this bug together

src/turtle_kv/change_log_file.cpp
tonyastolfi: Interesting; while I think the Chromium devs do make some good points, some of the examples seem a bit contrived (e.g., `emplace_back`-ing to a `std::vector<std::vector<int>>`)... The Google guidelines are a bit more cautious in some areas than I tend to feel is necessary, but I don't have a very strong feeling about it.

It's probably not likely to make a huge difference in this case, so I'll resolve the thread.

src/turtle_kv/kv_store.test.cpp
```diff
@@ -42,8 +44,130 @@ using turtle_kv::testing::run_workload;
...
+  void SetUp() override
+  {
+    auto root = turtle_kv::data_root();
+    ASSERT_TRUE(root.ok());
+    this->data_root = *root;
+
+    this->rng = std::default_random_engine{/*seed=*/1};
+    this->generate_key = RandomStringGenerator{};
+
+    SetupDefaultConfig();
```
tonyastolfi: nit: `this->`

src/turtle_kv/kv_store.test.cpp
```diff
@@ -42,8 +44,130 @@ using turtle_kv::testing::run_workload;
...
+      }
+
+      VLOG(3) << "Put key==" << key << ", value==" << value;
+    }
+  }
+
+  void ShutdownKVStore(std::unique_ptr<KVStore>& kv_store)
+  {
+    if (kv_store) {
+      std::this_thread::sleep_for(std::chrono::seconds(1));
```
tonyastolfi: Why is this needed?

src/turtle_kv/kv_store.test.cpp
```diff
@@ -473,14 +510,61 @@ TEST_P(CheckpointTest, CheckpointRecovery)
...
+      (*change_log_file)->read_blocks_into_vector();
+
+  ASSERT_TRUE(blocks.ok()) << BATT_INSPECT(blocks.status());
+
+  int i = 0;
+  for (auto block : *blocks) {
+    // TODO: [Gabe Bornstein 2/4/26] Investigate why verify is failing. It fails as soon as
+    // ChangeLogBlock is recovered.
+    //
+    // ASSERT_TRUE(block->verify().ok());
```
tonyastolfi: Can this be re-enabled?

src/turtle_kv/change_log_block.cpp
tonyastolfi: "leftovers from old debugging" -- Yes.

src/turtle_kv/change_log_block.hpp
tonyastolfi: Not pushed yet?

gabrielbornstein: Something weird is going on... maybe a UI issue? I can see the const added on the most recent commit here:
https://github.com/gabrielbornstein/turtle_kv/blob/gbornste/recover-memtable/src/turtle_kv/change_log_block.hpp
As well as in my branch, but it isn't showing up in the MR. Maybe it'll get fixed with my next commit? Either way, it should actually be there

src/turtle_kv/kv_store.test.cpp
gabrielbornstein: It's needed to test CheckpointRecovery. We sleep to wait for edits to be flushed and the checkpoint to be finalized because we're missing fsync. I'll move it from ShutDownKvStore and into the CheckpointRecovery test itself.

src/turtle_kv/kv_store.test.cpp
gabrielbornstein: Done

src/turtle_kv/kv_store.test.cpp
gabrielbornstein: Done

src/turtle_kv/kv_store.test.cpp
gabrielbornstein: Done

src/turtle_kv/change_log_block.cpp
```diff
@@ -11,26 +13,78 @@ namespace turtle_kv {
...
+  ChangeLogBlock* buffer = new (data) ChangeLogBlock{owner_id, std::move(grant), n_bytes};
 
   return buffer;
 }
 
+//==#==========+==+=+=++=+++++++++++-+-+--+----- --- -- -  -  -   -
+//
+/*static*/ StatusOr<boost::intrusive_ptr<ChangeLogBlock>> ChangeLogBlock::recover(
+    ChangeLogBlock::ScopedMemory&& memory,
+    batt::Grant&& grant)
```
tonyastolfi, line 45: In this case, I think the `memory` parameter should either be passed by value, or it should be `std::move`-ed into a local variable of type `ChangeLogBlock::ChangeLogBlockMemory`.  This is to make sure that the `memory` object that is passed in (or the last "live" copy of it) is actually destructed when `recover` returns.

src/turtle_kv/change_log_block.hpp
tonyastolfi: Problem solved; the `const` should be after the param list.

PR #11, opened 2026-02-11
Big MemTables support

tonyastolfi:
And some additional metrics collection in ResultSet.


src/turtle_kv/kv_store.cpp
```diff
@@ -573,8 +588,11 @@ Status KVStore::put(const KeyView& key, const ValueView& value) noexcept /*overr
 
     // Limit the number of deltas that can build up.
     //
-    BATT_REQUIRE_OK(this->deltas_size_->await_true([this](usize n) {
-      return n <= this->checkpoint_distance_.load() * 2;
+    const usize limit =
```
gabrielbornstein: Consider updating comment to include the units of limit. I'm assuming it's deltas, but I'd be surprised that the default # deltas when using big mem tables is 2.

src/turtle_kv/kv_store.cpp
```diff
@@ -573,8 +588,11 @@ Status KVStore::put(const KeyView& key, const ValueView& value) noexcept /*overr
...
+        this->runtime_options_.use_big_mem_tables ? 2 : (this->checkpoint_distance_.load() * 2);
```
gabrielbornstein, line 592: Isn't max(checkpoint_distance) == sizeof(memtable) in the worst case scenario? Does it make sense for the limit to be twice the size of a memtable?

src/turtle_kv/mem_table.hpp
```diff
@@ -112,11 +143,17 @@ class MemTable : public batt::RefCounted<MemTable>
     return id - 0x10000;
   }
 
+  static u64 ordinal_from_id(u64 id)
```
gabrielbornstein, line 156: Please add a comment explaining this function. I don't know what ordinal means.

src/turtle_kv/kv_store.cpp
```diff
@@ -881,22 +892,29 @@ Status KVStore::update_checkpoint(const State* observed_state)
...
-    if (checkpoint_job) {
-      BATT_REQUIRE_OK(this->commit_checkpoint(std::move(checkpoint_job)));
-    }
+    BATT_REQUIRE_OK(this->compact_memtable(  //
+        std::move(old_mem_table),
+        [this](std::unique_ptr<DeltaBatch> delta_batch) -> Status {
+          BATT_ASSIGN_OK_RESULT(std::unique_ptr<CheckpointJob> checkpoint_job,
+                                this->apply_batch_to_checkpoint(std::move(delta_batch)));
+
+          if (checkpoint_job) {
```
gabrielbornstein, line 907: Would it make more sense for apply_batch_to_checkpoint to return a Status? Looks like we're checking if it returns nullptr or not to determine if apply_batch_to_checkpoint was successful.

src/turtle_kv/kv_store.cpp
```diff
@@ -968,6 +986,10 @@ void KVStore::info_task_main() noexcept
 //
 void KVStore::memtable_compact_thread_main(usize thread_i)
 {
+  if (TURTLE_KV_BIG_MEM_TABLES) {
```
gabrielbornstein: Seems like we have two ways of determining if big_mem_tables are on, this global and this->runtime_options_.use_big_mem_tables. I'd suggest we use one or the other.

src/turtle_kv/kv_store.cpp
```diff
@@ -979,13 +1001,17 @@ void KVStore::memtable_compact_thread_main(usize thread_i)
       const u64 this_delta_batch_id = (**mem_table).id();
       const u64 next_delta_batch_id = MemTable::next_id_for(this_delta_batch_id);
 
-      std::unique_ptr<DeltaBatch> delta_batch = this->compact_memtable(std::move(*mem_table));
+      BATT_REQUIRE_OK(this->compact_memtable(
+          std::move(*mem_table),
+          [this, this_delta_batch_id](std::unique_ptr<DeltaBatch> delta_batch) -> Status {
+            while (this->next_delta_batch_to_push_.load() != this_delta_batch_id) {
+              batt::spin_yield();
```
gabrielbornstein, line 1012: Not sure exactly what spin_yield does but this is probably fine. Alternative design could be to have a mutex here and wherever else we are waiting for work to be finished. Once we acquire the mutex here, we could immediately drop it, just using it to determine when work is finished. Might take up less CPU if we're just spinning here right now. Or we send an alert to where we want to continue execution.

gabrielbornstein (commented):
Part 1 of my review, just going over the kv_store files. I'll look at the mem_table files next.

src/turtle_kv/mem_table.hpp
```diff
@@ -271,11 +354,29 @@ class MemTable : public batt::RefCounted<MemTable>
...
+#else  // TURTLE_KV_BIG_MEM_TABLES
+
   std::atomic<u32> compaction_state_{0};
 
   MergeCompactor::ResultSet</*decay_to_items=*/false> compacted_edits_;
 
-  batt::SmallVec<llfs::PageCache::ExternalAllocation, 16> cache_allocs_;
+#endif  // TURTLE_KV_BIG_MEM_TABLES
+
+  usize block_size_total_ = 0;
```
gabrielbornstein, line 390: Assuming this is bytes, but please specify units somewhere

src/turtle_kv/mem_table.hpp
```diff
@@ -271,11 +354,29 @@ class MemTable : public batt::RefCounted<MemTable>
...
+
+  i64 art_reserved_size_ = 0;
```
gabrielbornstein, line 394: Same as above, assuming this is bytes, but please specify units somewhere

src/turtle_kv/mem_table.hpp
```diff
@@ -271,11 +354,29 @@ class MemTable : public batt::RefCounted<MemTable>
...
+
+  usize since_last_cache_alloc_update_ = 0;
+
+  usize block_size_last_update_ = 0;
```
gabrielbornstein, line 404: Same as above, assuming this is bytes, but please specify units somewhere, general comment and trend I'd like to see more in this code base.

src/turtle_kv/mem_table.hpp
```diff
@@ -308,4 +417,49 @@ void MemTable::StorageImpl::store_data(usize n_bytes, SerializeFn&& serialize_fn
...
+
+#if TURTLE_KV_BIG_MEM_TABLES
+
+//=#=#==#==#===============+=+=+=+=++=++++++++++++++-++-+--+-+----+---------------
+/** \brief
+ */
+class MemTable::BatchCompactor
+{
+ public:
+  using Self = BatchCompactor;
```
gabrielbornstein, line 473: Do we give the definition of a Batch somewhere? For example, how many edits are in a batch, what's the min/max size of a batch, how does a Batch relate to blocks? I know a batch contains edits. but based on the comments, I don't know much more than that

src/turtle_kv/mem_table.hpp
```diff
@@ -226,24 +295,32 @@ class MemTable : public batt::RefCounted<MemTable>
...
 
   std::vector<EditView> compact_hash_index();
 
   std::vector<EditView> compact_art_index();
 
+#endif  // !TURTLE_KV_BIG_MEM_TABLES
+
+  i64 calculate_max_byte_size() const;
+
+  i64 update_external_cache_alloc();
```
gabrielbornstein, line 326: Add a doxygen style comment for this function

src/turtle_kv/mem_table.cpp
```diff
@@ -425,4 +425,181 @@ Slice<const EditView> MemTable::compacted_edits_slice_impl() const
   return {};
 }
 
+#endif  // !TURTLE_KV_BIG_MEM_TABLES
+
+//==#==========+==+=+=++=+++++++++++-+-+--+----- --- -- -  -  -   -
+//
+i64 MemTable::update_external_cache_alloc()
+{
+  i64 cache_alloc_delta = 0;
```
gabrielbornstein, line 437: Not sure what this variable name means... I'm guessing how much new data in bytes has been allocated to the cache? Could use a different name or a comment

src/turtle_kv/mem_table.hpp
```diff
@@ -271,11 +354,29 @@ class MemTable : public batt::RefCounted<MemTable>
...
   MergeCompactor::ResultSet</*decay_to_items=*/false> compacted_edits_;
 
-  batt::SmallVec<llfs::PageCache::ExternalAllocation, 16> cache_allocs_;
+#endif  // TURTLE_KV_BIG_MEM_TABLES
+
+  usize block_size_total_ = 0;
+
+  i64 art_reserved_size_ = 0;
+
+  usize since_last_cache_alloc_update_ = 0;
```
gabrielbornstein, line 399: What are this variables units? Just a monotonic counter?

src/turtle_kv/mem_table.cpp
```diff
@@ -425,4 +425,181 @@ Slice<const EditView> MemTable::compacted_edits_slice_impl() const
...
+
+  ++this->since_last_cache_alloc_update_;
+
+  if (!this->cache_alloc_in_progress_ && this->since_last_cache_alloc_update_ >= 128) {
```
gabrielbornstein: 128 is a magic number here, can we make it a const variable coming from somewhere?

gabrielbornstein (commented):
Second round, I'll finish out my review of mem_table.cpp tomorrow

src/turtle_kv/kv_store.cpp
tonyastolfi: Renamed to `max_deltas_size`.

src/turtle_kv/kv_store.cpp
tonyastolfi: Prior to introducing Big MemTables, yes the limit was equal to the checkpoint distance (which could be thought of as either being in units of MemTable, batch, or leaf page).  Now, there is a single MemTable that grows to include the entire checkpoint distance.  Checkpoint distance is now given in units of batch/leaf.

The rationale for `2` (previously `checkpoint_distance * 2`) is to allow for parallel processing of two stages of the checkpoint update pipeline downstream of finalized MemTables:

1. Building the next checkpoint (turtle) tree in memory (the CheckpointGenerator)
2. Writing the new pages for a finalized Checkpoint to storage

src/turtle_kv/mem_table.hpp
tonyastolfi: Added doc comment to the function.

src/turtle_kv/kv_store.cpp
tonyastolfi: Yeah, that's a fair comment.  

With the Big MemTable design, it doesn't really make sense to have multiple memtable compaction threads, since there isn't enough work at that pipeline stage to parallelize, so I think this code won't be very long-lived.  I've added a TODO comment to remove it once we don't have to worry about being the right thread to push to the checkpoint update channel next.

src/turtle_kv/mem_table.hpp
tonyastolfi: Added comment.

src/turtle_kv/mem_table.cpp
tonyastolfi: Commented.

src/turtle_kv/mem_table.hpp
tonyastolfi: It's the unit of update for a TurtleTree (as in https://github.com/mathworks/turtle_kv/blob/main/src/turtle_kv/tree/subtree.hpp#L99).  Don't know if a formal definition appears per se.

src/turtle_kv/mem_table.hpp
tonyastolfi: Done.

src/turtle_kv/mem_table.cpp
tonyastolfi: Added named constant `MemTable::kBlocksPerExternalCacheAllocUpdate` for this value.  See the doc comment in `mem_table.hpp` for a description of what it does.

src/turtle_kv/mem_table.hpp
tonyastolfi: Done.

src/turtle_kv/mem_table.hpp
tonyastolfi: Commented.

src/turtle_kv/mem_table.hpp
tonyastolfi: Commented.

src/turtle_kv/kv_store.cpp
tonyastolfi: `apply_batch_to_checkpoint` is successful whether or not it returns `nullptr`.  The difference is whether the passed batch was determined to be the last one in the checkpoint, in which case extra work is done to create a `CheckpointJob` which durably commits the new pages.

tonyastolfi: The `BATT_ASSIGN_OK_RESULT` is already unwrapping a `StatusOr`, which is what `apply_batch_to_checkpoint` returns.

src/turtle_kv/kv_store.cpp
tonyastolfi: I agree that this isn't the best way to write this.  

The macro `TURTLE_KV_BIG_MEM_TABLES` controls whether support for big memtables is compiled in.  The runtime option `use_big_mem_tables` can be set to true to false to change the behavior at runtime.  

This enables A/B comparison of how big memtables changes performance for different workloads, without requiring a compilation in between.  

I've change this to be conditional compilation instead of a regular `if`.

src/turtle_kv/kv_store.cpp
```diff
@@ -1197,9 +1308,36 @@ Status KVStore::commit_checkpoint(std::unique_ptr<CheckpointJob>&& checkpoint_jo
...
-                                old_state->deltas_.end());
+      {
+#if TURTLE_KV_BIG_MEM_TABLES
+
+        // Find the first delta MemTable on the stack that is *not* covered by the checkpoint.
+        //  (deltas_ is in oldest-to-newest order)
+        //  TODO [tastolfi 2025-11-22] binary search?
+        //
+        auto iter = old_state->deltas_.begin();
+        while (iter != old_state->deltas_.end()) {
```
vidyasilai: As the TODO above mentions, I think we could use binary search here?

src/turtle_kv/kv_store_scanner.cpp
```diff
@@ -7,6 +7,27 @@ namespace turtle_kv {
...
+  return scanner.get_key();
+}
+
+//==#==========+==+=+=++=+++++++++++-+-+--+----- --- -- -  -  -   -
+//
+template <ARTBase::Synchronized kSynchronized>
+KeyView art_scanner_get_key(ART<MemTableValueEntry>::Scanner<kSynchronized,
+                                                             /*kValuesOnly=*/true>& scanner)
+{
+  return scanner.get_value().key_view();
```
vidyasilai, line 26: For my own understanding, does `scanner.get_value().key_view()` and `scanner.get_value().value_view()` return the same thing in the values only case?

src/turtle_kv/core/merge_compactor.hpp
```diff
@@ -58,6 +58,9 @@ class MergeCompactor : public MergeCompactorBase
     StatsMetric<i64> result_set_waste;
     CountMetric<usize> result_set_compact_count;
     CountMetric<usize> result_set_compact_byte_count;
+    CountMetric<i64> result_set_bytes_alloc;
```
vidyasilai, line 61: Should we be updating this metric in the `concat` function?

vidyasilai (commented):
Everything looks great to me! Just a few small questions/comments.

src/turtle_kv/core/merge_compactor.hpp
tonyastolfi: I can see the argument for doing this; we would also be updating `result_set_bytes_freed` metric by the same amount, however, since `concat` consumes the two input sets.  So the fact that we don't update either in `concat` does preserve the estimated total footprint of all result sets (`result_set_bytes_alloc - result_set_bytes_freed`).  

The only thing `concat` is really allocating are the much smaller vectors to hold shared pointers to the vectors of edits plus the active range views.   I don't think its worth tracking these.

src/turtle_kv/kv_store_scanner.cpp
tonyastolfi: No, not exactly.  A "value" from the ARTScanner's perspective is a MemTableValueEntry, which itself has a member function called `value_view()` which returns the actual edit's value (a ValueView).  Because MemTableValueEntry also has a `key_view()`, it would be wasted work to build keys using the ART node paths inside the scanner.  

What is true is that if you have a non-values-only scanner `kv_scanner` and a values-only scanner `v_scanner`, and the ART "value" type is MemTableValueEntry in both cases, then `kv_scanner.get_key()` will always equal `kv_scanner.get_value().key_view()`; and also `kv_scanner.get_value()` will always equal `v_scanner.get_value()` when they are iterated in sync with each other.

src/turtle_kv/kv_store.cpp
tonyastolfi: Fixed.

src/turtle_kv/kv_store_scanner.cpp
vidyasilai: Thanks for the explanation!

vidyasilai (approved):

PR #14, opened 2026-03-10
Add active pivot overflow bit set to Update Buffer Segment

vidyasilai:
Adds another 64-bit bit set to `InMemoryNode::UpdateBuffer::Segment` in order to handle the intermediate steps involved in node merging, during which pivot count could exceed 64.

### DIFF
--- a/src/turtle_kv/tree/packed_node_page.hpp
+++ b/src/turtle_kv/tree/packed_node_page.hpp
@@ -49,7 +49,7 @@ struct PackedNodePage {
       kMaxPivots + 1 /*max_key*/ + 1 /*common_prefix*/ + 1 /*final_offset*/;
 
   static constexpr u8 kFlagSizeTiered = 0x80;
-  static constexpr u8 kPivotCountMask = 0x3f;
+  static constexpr u8 kPivotCountMask = 0x7f;
   static constexpr u16 kSegmentStartsFiltered = 0x8000;
 
   using Key = PackedNodePageKey;
@@ -141,6 +141,11 @@ struct PackedNodePage {
       u64 get_active_pivots() const
       {
         return this->active_pivots;
+      }
+
+      std::pair<u64, u64> get_active_pivots_with_overflow() const
+      {
+        return std::make_pair(this->active_pivots, u64{0});
       }
 
       llfs::PageId get_leaf_page_id() const
--- a/src/turtle_kv/tree/packed_node_page.hpp
+++ b/src/turtle_kv/tree/packed_node_page.hpp
@@ -143,13 +143,6 @@ struct PackedNodePage {
         return this->active_pivots;
       }
 
-#if 0
-      std::array<u64, 2> get_active_pivots_with_overflow() const
-      {
-        return {this->active_pivots, u64{0}};
-      }
-#endif
-
       llfs::PageId get_leaf_page_id() const
       {
         return this->leaf_page_id.unpack();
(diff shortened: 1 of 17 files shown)


src/turtle_kv/tree/testing/fake_segment.hpp
```diff
@@ -84,40 +86,51 @@ struct FakeSegment {
...
+    this->active_pivots_ = {};
     this->pivot_items_count_.clear();
   }
 
+  bool is_inactive() const
+  {
+    const bool inactive = this->active_pivots_.is_empty();
+    if (inactive) {
+      Slice<const Interval<u32>> filter_dropped_ranges = this->filter_.dropped();
+      BATT_CHECK_EQ(filter_dropped_ranges.size(), 1);
```
tonyastolfi, line 98: +1: Nice defensive checks in this function!

src/turtle_kv/tree/packed_node_page.hpp
```diff
@@ -42,14 +43,13 @@ struct Subtree;
...
 struct PackedNodePage {
-  static constexpr usize kMaxPivots = 64;
   static constexpr usize kMaxSegments = kMaxPivots - 1;
   static constexpr usize kMaxLevels = batt::log2_ceil(kMaxPivots);
   static constexpr usize kPivotKeysSize =
       kMaxPivots + 1 /*max_key*/ + 1 /*common_prefix*/ + 1 /*final_offset*/;
 
   static constexpr u8 kFlagSizeTiered = 0x80;
-  static constexpr u8 kPivotCountMask = 0x3f;
+  static constexpr u8 kPivotCountMask = 0x7f;
```
tonyastolfi: This needs to be at least 64 right?  It does bug me that we can't reserve 6 bits for this anymore.  Adding comment at least...

src/turtle_kv/util/piecewise_filter.test.cpp
```diff
@@ -63,68 +64,82 @@ TEST(PiecewiseFilterTest, InvalidFilterTest)
...
 
-    // First verify that everything should be live since nothing has been dropped yet.
+    // All items start unfiltered.
     //
-    EXPECT_EQ(filter.dropped_total(), 0);
-
-    Interval<usize> next_live_interval = filter.find_live_range(Interval<usize>{0, num_items});
-    EXPECT_EQ(next_live_interval, (Interval<usize>{0, num_items}));
+    std::set<usize> live_items;
+    for (usize i = 0; i < num_items; ++i) {
```
tonyastolfi, line 78: It would be better if the inner variable name didn't shadow the top-level loop.  I suggest renaming the top-level one to `seed`.

src/turtle_kv/util/piecewise_filter.test.cpp
```diff
@@ -63,68 +64,82 @@ TEST(PiecewiseFilterTest, InvalidFilterTest)
 //
 TEST(PiecewiseFilterTest, QueryTest)
 {
+  const usize num_items = 1000000;
```
tonyastolfi: I think we'd probably get better coverage if `num_items` were smaller.  Maybe 100 or 1000 max.

src/turtle_kv/util/piecewise_filter.test.cpp
vidyasilai: Fixed

src/turtle_kv/util/piecewise_filter.test.cpp
vidyasilai: Fixed, `num_items` is now 1000.

PR #17, opened 2026-03-18
Fix for #16.

### LINKED ISSUE #16: Sporadic panic on KVStore shutdown
Found in test `KVStoreTest.CreateAndOpen`:

```
[----------] Global test environment set-up.
[----------] 1 test from KVStoreTest
[ RUN      ] KVStoreTest.CreateAndOpen
WARNING: Logging before InitGoogleLogging() is written to STDERR
I20260318 17:26:45.784240 139766592960064 kv_store.test.cpp:75]  tree_options.filter_bits_per_key() == 12 tree_options.filter_page_size() == 131072
I20260318 17:26:45.811544 139766592960064 page_cache_slot_pool.cpp:64] PageCacheSlot::Pool created, n_slots=1048576

...

E20260318 17:26:50.261707 139766592960064 page_cache_slot.cpp:27] FATAL: /home/tastolfi/github/runners/llfs/_work/llfs/llfs/src/llfs/page_cache_slot.cpp:27: Assertion failed: bool{!this->is_pinned()} == true
 (in `llfs::PageCacheSlot::~PageCacheSlot()`)

  bool{!this->is_pinned()} == 0

  true == 1

 this->pin_count() == 1 this->cache_slot_ref_count() == 0 (void*)this == 0x7f1d4ffff440
E20260318 17:26:50.261722 139766592960064 assert_impl.hpp:19]

build/Release/src/turtle_kv_Test(+0x8b410)[0x55beed9df410]
build/Release/src/turtle_kv_Test(+0x45bb11)[0x55beeddafb11]
build/Release/src/turtle_kv_Test(+0x1db373)[0x55beedb2f373]
build/Release/src/turtle_kv_Test(+0x1c2a0f)[0x55beedb16a0f]
build/Release/src/turtle_kv_Test(+0x1d032e)[0x55beedb2432e]
build/Release/src/turtle_kv_Test(+0x922b3)[0x55beed9e62b3]
build/Release/src/turtle_kv_Test(+0xd8a92)[0x55beeda2ca92]
/lib/x86_64-linux-gnu/libstdc++.so.6(+0xe62b3)[0x7f1df26e62b3]
/lib/x86_64-linux-gnu/libc.so.6(+0x94ac3)[0x7f1df2294ac3]
/lib/x86_64-linux-gnu/libc.so.6(+0x126850)[0x7f1df2326850]

 0# batt::print_stack_trace() in build/Release/src/turtle_kv_Test
 1# batt::fail_check_exit() in build/Release/src/turtle_kv_Test
 2# llfs::PageCacheSlot::Pool::~Pool() in build/Release/src/turtle_kv_Test
 3# llfs::PageCache::~PageCache() in build/Release/src/turtle_kv_Test
 4# std::_Sp_counted_ptr<llfs::PageCache*, (__gnu_cxx::_Lock_policy)2>::_M_dispose() in build/Release/src/turtle_kv_Test
 5# void boost::sp_adl_block::intrusive_ptr_release<llfs::StorageContext, boost::sp_adl_block::thread_safe_counter>(boost::sp_adl_block::intrusive_ref_counter<llfs::StorageContext, boost::sp_adl_block::thread_safe_counter> const*) in build/Release/src/turtle_kv_Test
 6# (anonymous namespace)::KVStoreTest_CreateAndOpen_Test::TestBody()::{lambda()#1}::operator()() const in build/Release/src/turtle_kv_Test
 7# 0x00000000000E62B3 in /lib/x86_64-linux-gnu/libstdc++.so.6
 8# start_thread at ./nptl/pthread_create.c:442
 9# __GI___clone3 at ../sysdeps/unix/sysv/linux/x86_64/clone3.S:83

 0. Destructing slot 8 n_to_delete == 19 this->n_allocated_.load() == 19 this->n_slots_ == 1048576
    at llfs/page_cache_slot_pool.cpp:80
    in llfs::PageCacheSlot::Pool::~Pool()::<lambda(std::ostream&, const void*)>
    stack offset: --
FATAL: signal 6 (Aborted):
[[raw stack]]
```

tonyastolfi:
Fixes #16

### DIFF
--- a/src/turtle_kv/kv_store.hpp
+++ b/src/turtle_kv/kv_store.hpp
@@ -1,4 +1,8 @@
 #pragma once
+#define TURTLE_KV_KV_STORE_HPP
+
+#include <turtle_kv/config.hpp>
+//
 
 #include <turtle_kv/change_log_writer.hpp>
 #include <turtle_kv/checkpoint.hpp>
@@ -7,6 +11,7 @@
 #include <turtle_kv/kv_store_metrics.hpp>
 #include <turtle_kv/mem_table.hpp>
 
+#include <turtle_kv/tree/filter_page_write_state.hpp>
 #include <turtle_kv/tree/pinning_page_loader.hpp>
 #include <turtle_kv/tree/tree_options.hpp>
 
@@ -253,6 +258,8 @@ class KVStore : public Table
   absl::Mutex base_checkpoint_mutex_;
 
   std::unique_ptr<llfs::Volume> checkpoint_log_;
+
+  boost::intrusive_ptr<FilterPageWriteState> filter_page_write_state_;
 
   ObjectThreadStorage<KVStore::ThreadContext>::ScopedSlot per_thread_;
 
--- a/src/turtle_kv/tree/filter_page_write_state.cpp
+++ b/src/turtle_kv/tree/filter_page_write_state.cpp
@@ -1,0 +1,81 @@
+#include <turtle_kv/tree/filter_page_write_state.hpp>
+//
+#include <glog/logging.h>
+
+namespace turtle_kv {
+
+//==#==========+==+=+=++=+++++++++++-+-+--+----- --- -- -  -  -   -
+//
+/*static*/ auto FilterPageWriteState::make_new() noexcept -> boost::intrusive_ptr<Self>
+{
+  return boost::intrusive_ptr<Self>{new Self{}};
+}
+
+//==#==========+==+=+=++=+++++++++++-+-+--+----- --- -- -  -  -   -
+//
+/*explicit*/ FilterPageWriteState::FilterPageWriteState() noexcept
+{
+}
+
+//==#==========+==+=+=++=+++++++++++-+-+--+----- --- -- -  -  -   -
+//
+i64 FilterPageWriteState::active_count() const
+{
+  const i64 observed_finished = this->finished_count();
+  const i64 observed_started = this->started_count();
+
+  return observed_started - observed_finished;
+}
+
+//==#==========+==+=+=++=+++++++++++-+-+--+----- --- -- -  -  -   -
+//
+void FilterPageWriteState::halt()
+{
+  this->started_.fetch_or(1);
+  this->finished_.poke();
+}
+
+//==#==========+==+=+=++=+++++++++++-+-+--+----- --- -- -  -  -   -
+//
+void FilterPageWriteState::join()
+{
+  for (;;) {
+    Status status = this->finished_
+                        .await_true([this](i64 observed_finished) {
+                          const i64 observed_started = this->started_.load();
+                          return observed_started == observed_finished * 2 + 1;
+                        })
+                        .status();
+    if (status != batt::StatusCode::kPoke) {
+      break;
+    }
+  }
+}
+
+//==#==========+==+=+=++=+++++++++++-+-+--+----- --- -- -  -  -   -
+//
+Status FilterPageWriteState::start()
+{
+  const i64 observed = this->started_.fetch_add(2);
+  if (Self::is_halted_state(observed)) {
+    // Undo the change to started_.
+    //
+    this->started_.fetch_sub(2);
+    this->finished_.poke();
+    return batt::StatusCode::kClosed;
+  }
+  return OkStatus();
+}
+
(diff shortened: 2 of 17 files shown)


src/turtle_kv/tree/filter_page_write_state.hpp
```diff
@@ -0,0 +1,115 @@
...
+ */
+class FilterPageWriteState : public batt::RefCounted<FilterPageWriteState>
+{
+ public:
+  using Self = FilterPageWriteState;
+
+  /** \brief Returns true iff the observed value of the `started_` counter indicates that halt has
+   * been called.
+   */
+  static constexpr bool is_halted_state(i64 observed_started)
```
gabrielbornstein, line 31: Consider making this private. Looks like it would only get called by member functions.

src/turtle_kv/tree/filter_page_write_state.hpp
```diff
@@ -0,0 +1,115 @@
...
+   */
+  explicit FilterPageWriteState() noexcept;
+
+  //+++++++++++-+-+--+----- --- -- -  -  -   -
+
+  /** \brief Holds the number of successful starts plus the halted state.
+   */
+  std::atomic<i64> started_{0};
+
+  /** \brief Tracks the number of finishes.
```
gabrielbornstein, line 110: Good comments throughout this file! Feel free to resolve this comment.

src/turtle_kv/tree/filter_page_write_state.cpp
```diff
@@ -0,0 +1,81 @@
...
+      break;
+    }
+  }
+}
+
+//==#==========+==+=+=++=+++++++++++-+-+--+----- --- -- -  -  -   -
+//
+Status FilterPageWriteState::start()
+{
+  const i64 observed = this->started_.fetch_add(2);
```
gabrielbornstein, line 59: Why are we fetch adding and fetch subbing two instead of one?

src/turtle_kv/kv_store.hpp
```diff
@@ -254,6 +259,8 @@ class KVStore : public Table
 
   std::unique_ptr<llfs::Volume> checkpoint_log_;
 
+  boost::intrusive_ptr<FilterPageWriteState> filter_page_write_state_;
```
gabrielbornstein, line 262: Consider using shared_ptr instead of intrusive_ptr. Is there a reason we need to be managing increments and decrements ourselves? If not, let's let shared_ptr do it for us.

Also, maybe I just missed the definitions, but I don't see add_ref or remove_ref for FilterPageWriteState defined anywhere.

gabrielbornstein (commented):
Some small comments. Changes look great, just reply to my feedback/questions and we can merge.

src/turtle_kv/tree/filter_page_write_state.hpp
tonyastolfi: Thanks!

src/turtle_kv/kv_store.hpp
tonyastolfi: We don't need to define the add/remove ref functions in this case because `FilterPageWriteState` derives from `batt::RefCounted`.  It's reasonable to consider using `std::shared_ptr` instead, but for this case, I think using a self-contained ref count offers the following advantages:

1. This class is only ever managed by ref count; so the intrusive counter isn't ever wasted
2. `boost::intrusive_ptr` is the size of a single pointer, whereas `std::shared_ptr` is two pointers (one for the managed object, one for the ref counter)

The second point seems pretty minor, but given that a (smart) ptr to this object is captured in the (lambda) handler for the asynchronous write of the filter page(s), it is worth doing.  When lambdas, or any callback type, is stored in a type-erased container like `std::function`, whether or not the container needs to allocate memory to store the value is determined by how big the value is.  `std::function` (and other type-erasers) typically have a small-object buffer internally which allows them to avoid hitting the heap allocator if the callable value you're storing/passing is small enough.

src/turtle_kv/tree/filter_page_write_state.cpp
tonyastolfi: Because the least-significant bit is used as a flag storing whether the object has been "halted."  That's so that we don't keep accumulating more filter page writes once the KVStore has started shutting down.

PR #20, opened 2026-04-15
Update Segment Filters to Use Live Intervals

vidyasilai:


src/turtle_kv/util/piecewise_filter.ipp
```diff
@@ -63,74 +81,134 @@ bool PiecewiseFilter<OffsetT>::check_invariants() const
...
+                                        typename Interval<OffsetT>::LinearOrder{});
+
+  // Edge case: if dropping a sub-range of an already dropped range, do nothing.
   //
-  auto iter = std::lower_bound(this->dropped_.begin(),
-                               this->dropped_.end(),
-                               i,
-                               typename Interval<OffsetT>::LinearOrder{});
+  if (first == last) {
+    Interval<OffsetT> dropped;
```
tonyastolfi: Let's always initialize `dropped` with the code in this block.

src/turtle_kv/util/piecewise_filter.ipp
```diff
@@ -63,74 +81,134 @@ bool PiecewiseFilter<OffsetT>::check_invariants() const
...
   //
-  if (iter != this->dropped_.begin()) {
-    auto before = std::prev(iter);
-    if (iter->adjacent_to(*before)) {
-      this->dropped_total_ -= before->size() + iter->size();
-      *before = before->union_with(*iter);
-      this->dropped_total_ += before->size();
-      this->dropped_.erase(iter);
-      iter = before;
+  while (first != this->live_.end()) {
```
tonyastolfi, line 167: we can have `while (first != last) {` and then the `break` condition immediately below is covered (so can be removed)

src/turtle_kv/util/piecewise_filter.ipp
```diff
@@ -146,18 +224,37 @@ bool PiecewiseFilter<OffsetT>::live_at_index(OffsetT i) const
...
+  auto iter = std::lower_bound(this->live_.begin(),
+                               this->live_.end(),
                                i,
                                typename Interval<OffsetT>::LinearOrder{});
 
-  if (iter != this->dropped_.end() && iter->contains(i)) {
-    return iter->upper_bound;
+  // Check if the previous interval contains `i`.
+  //
+  if (iter != this->live_.begin()) {
```
tonyastolfi: Double check to see if this case is ever hit; I'm thinking probably it isn't.

src/turtle_kv/util/piecewise_filter.ipp
```diff
@@ -170,36 +267,120 @@ Interval<OffsetT> PiecewiseFilter<OffsetT>::find_live_range(Interval<OffsetT> i)
...
+  auto iter = std::lower_bound(this->live_.begin(),
+                               this->live_.end(),
                                start_i,
                                typename Interval<OffsetT>::LinearOrder{});
 
-  // Start by finding the live lower bound of start_i. If start_i is filtered, adjust it to be the
-  // live lower bound.
+  // Check if the previous interval contains `start_i`.
+  //
+  if (iter != this->live_.begin()) {
```
tonyastolfi: This case may not be needed also (see comment in live lower bound)

src/turtle_kv/tree/packed_node_page.cpp
```diff
@@ -151,21 +151,19 @@ PackedNodePage* build_node_page(const MutableBuffer& buffer, const InMemoryNode&
...
-
-          // If the first item is filtered, set the most significant bit of `filter_start` to 1.
+        Slice<const Interval<u32>> live_ranges = segment_filter.live();
+        if (!live_ranges.empty()) {
+          // If the first item is live, set the most significant bit of `filter_start` to 1.
           //
-          bool start_filtered = dropped_ranges[0].lower_bound == 0;
-          if (start_filtered) {
-            dst_segment.filter_start |= PackedNodePage::kSegmentStartsFiltered;
+          bool start_live = live_ranges[0].lower_bound == PiecewiseFilter<u32>::kMinLowerBound;
```
tonyastolfi, line 158: Maybe also add special case for kMaxUpperBound where we avoid storing it?

vidyasilai: Done

src/turtle_kv/util/piecewise_filter.ipp
vidyasilai: Done

src/turtle_kv/util/piecewise_filter.ipp
vidyasilai: Done

src/turtle_kv/util/piecewise_filter.ipp
vidyasilai: After thinking about this again, I kept the original code. In Case 1 where we do `first = this->live_.erase(first);`, the call the `erase` will invalidate `last`. Let me know what you think and if there's a different way we could write this.

src/turtle_kv/util/piecewise_filter.ipp
vidyasilai: I refactored this code slightly, let me know what you think. I still think we need to keep the conditional checks `to_drop.lower_bound < first->lower_bound` and `to_drop.upper_bound >= std::prev(last)->upper_bound` in the cases where `to_drop` starts and/or ends in a dropped region, in order to return to the caller the maximal contiguous dropped region that `to_drop` lies in.

src/turtle_kv/tree/in_memory_node.hpp
```diff
@@ -160,22 +155,17 @@ struct InMemoryNode {
...
-        }
-        BATT_CHECK(!dropped_ranges.empty());
+        Slice<const Interval<u32>> live_ranges = this->filter.live();
 
-        // If the first element is filtered out, don't include 0 as a cut point, only include the
+        // If the first element is live, don't include 0 as a cut point, only include the
         // upper bound.
         //
-        bool first_element_filtered = dropped_ranges[0].lower_bound == 0;
+        bool first_element_live = live_ranges.empty() || live_ranges.front().lower_bound ==
```
tonyastolfi: Seems like this should be `!live_ranges.empty() && live_ranges.front().lower_bound == kMinLowerBound`?

src/turtle_kv/tree/in_memory_node.hpp
```diff
@@ -160,22 +155,17 @@ struct InMemoryNode {
...
+                                                             PiecewiseFilter<u32>::kMinLowerBound;
+        bool ends_at_max = live_ranges.empty() ||
```
tonyastolfi: Seems like this should be `!live_ranges.empty() && live_ranges.back().upper_bound == kMaxUpperBound`?

src/turtle_kv/util/piecewise_filter.ipp
tonyastolfi: That's a good point!  Ok, it's good as-is IMO

src/turtle_kv/tree/in_memory_node.hpp
vidyasilai: Fixed

src/turtle_kv/tree/in_memory_node.hpp
vidyasilai: Fixed

PR #26, opened 2026-06-22
Fixes for exporting to conan cache

vidyasilai:


test_package/conanfile.py
```diff
@@ -12,6 +12,7 @@ class turtle_kvTestConan(ConanFile):
 
     def requirements(self):
         self.requires(self.tested_reference_str)
+        self.requires("batteries/[>=0.71.1 <1]", force=True)
```
tonyastolfi: I think this should be marked as `visible=True`, instead of explicitly adding `batteries::batteries` in the downstream package's link libraries.

See https://docs.conan.io/2/reference/conanfile/methods/requirements.html#visible

vidyasilai: Fixed

